PR #68, opened 2022-02-09
Update vue and vue-router

msimecek:
There were breaking changes between Vue 2 and 3, which caused Vue Router to also break on upgrade. This PR fixes both.

Tested here: https://afe2e0a36-global-fd.azurefd.net/


sebader (approved):
thank you! please move to -connected as well

sebader:
nobody promised that this process would be fun ;)

PR #98, opened 2022-02-15
WIP - Mermaid experiments

heoelri:
This adds a mermaid based state diagram to the Health Service docs to potentially replace an existing image.

image

vs. 

image

### DIFF
--- a/src/app/AlwaysOn.HealthService/README.md
+++ b/src/app/AlwaysOn.HealthService/README.md
@@ -4,6 +4,25 @@ The HealthService is an app component that is running along other components lik
 
 ![HealthService conceptual diagram](/docs/media/health-service-high-level.png)
 
+```mermaid
+stateDiagram
+    direction LR
+    [*] --> StampHealthCheck
+    state FrontDoor {
+    StampHealthCheck --> HealthService
+    }
+    HealthService --> CosmosDB: Able to query?
+    HealthService --> Storage: File exists?
+    HealthService --> EventHubs: Able to sent message?
+    state AKS {
+        direction RL
+        Application --> HealthService: LivenessProbe
+        BackendProcess --> HealthService: LivenessProbe
+    }
+
+```
+
+
 The idea is, first of all, if the cluster itself is down, the health service won't respond at all. When the service is up and running, it performs periodic checks against various components of the solution:
 
 - It attempts to do a simple query against Cosmos DB
--- a/src/app/AlwaysOn.HealthService/README.md
+++ b/src/app/AlwaysOn.HealthService/README.md
@@ -11,17 +11,16 @@ stateDiagram
     state FrontDoor {
     StampHealthCheck --> HealthService
     }
+    state AKS {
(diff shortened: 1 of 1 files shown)


sebader:
the health service is actually not calling the Liveness Probes of the other Pods. But maybe it should (could?)?! :) thoughts @msimecek ?

heoelri:
> the health service is actually not calling the Liveness Probes of the other Pods. But maybe it should (could?)?! :) thoughts @msimecek Martin Simecek FTE ?

VSCode doesn't support mermaid, yet. That's why I've ended with the draft PR.. 

The diagram was supposed to visualize that Application and Backend Process representing deployments/pods use HealthService for their liveness probe.

heoelri:
@sebader this is what i was trying to visualize:

`/src/app/charts/catalogservice/templates/deployment.yaml`

```yaml
        livenessProbe:
          httpGet:
            path: /health/liveness
            port: {{ .Values.workload.port | default 8080 }}
          initialDelaySeconds: 10
          periodSeconds: 30
          timeoutSeconds: 3
        startupProbe:
          httpGet:
            path: /health/liveness
            port: {{ .Values.workload.port | default 8080 }}
```

But you're right, they're pointing to themselfs.

heoelri:
Its then probably more like this:

image

heoelri:
And to respond to the initial question - means that the healthservice would currently not recognize that the application deployment (or the background processor) does not work? @msimecek / @sebader

sebader:
I'm afraid to say that you made a very valid observation ;) We only check the dependencies (i.e. the Azure services) as of today

sebader:
> But you're right, they're pointing to themselfs.

well, technically it is k8s checking the pods

heoelri:
Drawing diagrams helps!

heoelri:
> > But you're right, they're pointing to themselfs.
> 
> well, technically it is k8s checking the pods

true, the kubelet on each node but not sure if that detail adds any value here

sebader:
> > > But you're right, they're pointing to themselfs.
> > 
> > 
> > well, technically it is k8s checking the pods
> 
> true, the kubelet on each node but not sure if that detail adds any value here

probably not. I'd leave it out

heoelri:
```mermaid
stateDiagram
    direction LR
    [*] --> FrontDoor: HTTPS
    FrontDoor --> IngressController: HTTPS

    state GlobalServices {
        FrontDoor
        FrontDoor --> Storage: Read Product Images
        CosmosDB
        Storage
    }
    state RegionalStamp {
        AzureKubernetesService
        EventHub
    }
    state AzureKubernetesService {
        direction RL
        IngressController
        IngressController --> CatalogService: HTTP
        CatalogService --> CoreDNS: DNS lookups
    }
    CatalogService --> CosmosDB: Query
    CatalogService --> EventHub: Write
    CatalogService --> Storage: Store Product Images
```

heoelri:
Closed - will be superseeded in a bit.

PR #101, opened 2022-02-15
Version information to the UI

msimecek:
Cosmetic but useful touch: adding the version label to the UI to be able to quickly tell which _variant_, _branch_ and _build_ we are looking at.

image

### DIFF
--- a/.ado/pipelines/config/configuration.yaml
+++ b/.ado/pipelines/config/configuration.yaml
@@ -1,6 +1,9 @@
 # Central configuration and versioning settings
 # used for all pipelines
 variables:
+- name: 'riVariant' # Reference implementation variant - each repository should have its own identification
+  value: 'foundational-online'
+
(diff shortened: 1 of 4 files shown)


.ado/pipelines/config/configuration.yaml
```diff
@@ -1,6 +1,9 @@
 # Central configuration and versioning settings
 # used for all pipelines
 variables:
+- name: 'riVariant' # Reference implementation variant - each repository should have its own identification
```
sebader, line 4: should we add a comment where this is being used?

msimecek: Reading my mind :) I was thinking about that, but then decided not to add it, in case this value is used somewhere else than in the UI. But since it isn't for now, I will add it.

sebader:
and off to -connected :)

PR #104, opened 2022-02-16
Extending HealthService to probe pods

heoelri:
@msimecek / @sebader following up on the conversation we had in #98 what do you guys think about adding 1, 2 or more additional `LIVENESS_ENDPOINT_x` (name is tdb) the healthservice can probe additionally to what it does today. 

This will of course not probe each individual pod - it will use the ClusterIP service instead.

```mermaid
stateDiagram
    direction LR
    [*] --> StampHealthCheck
    state FrontDoor {
    StampHealthCheck --> HealthService
    }
    state AKS {
        direction RL
        HealthService
        HealthService --> Application: Alive?
        HealthService --> BackendProcess: Alive?
    }
    HealthService --> CosmosDB: Able to query?
    HealthService --> Storage: File exists?
    HealthService --> EventHubs: Able to sent message?
```

Treat my implementation just a sample I've used to test that - you can probably implement it better.

### DIFF
--- a/src/app/AlwaysOn.HealthService/AlwaysOnHealthCheck.cs
+++ b/src/app/AlwaysOn.HealthService/AlwaysOnHealthCheck.cs
@@ -116,9 +116,9 @@ namespace AlwaysOn.HealthService
         private async Task<bool> GetLivenessEndpoint1Status(CancellationToken cancellationToken = default (CancellationToken))
         {
             var result = true;
-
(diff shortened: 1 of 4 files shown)


heoelri:
@sebader / @msimecek in accordance to our discussion in team I'm going to close this PR. We came to the conclusion that querying the k8s apiserver to check the health of the application is probably the best solution. Feel free to re-use my mermaid diagram.

PR #109, opened 2022-02-17
ADO provisioning script updates

msimecek:
- fixed missing comma
- added the option to provide existing service principal ID
  - the script then asks for SP password during the run


sebader (approved):
didnt test it again but looks right and assume you did test :)

PR #116, opened 2022-02-23
Grafana - Adding a default nsg for compliance

heoelri:
@nielsams adding a default NSG (similar to what we've in AlwaysOn for the regional stamps) for compliance reasons.


nielsams (approved):

sebader (approved):

heoelri:
Thanks! Just to make sure that we're all on the same page - it wasn't tested, yet!

heoelri:
Shouldn't break anything though - @nielsams there's no destroy option for grafana, right? Can i just run this against the existing int environment to test these changes before merging them into main?

heoelri:
Successfully tested.

image
https://dev.azure.com/AlwaysOn/Foundational-Online/_build/results?buildId=417&view=results

New NSG per stamp is getting created:
image

And associated to **both** subnets:
image

PR #145, opened 2022-02-23
check on az errors of blob upload

sebader:


sebader:
and on -Connected https://github.com/Azure/AlwaysOn-Foundational-Connected/pull/68

heoelri (approved):

PR #146, opened 2022-02-23
Component updates

msimecek:
Collecting and testing bunch of dependa updates from today.


msimecek:
https://dev.azure.com/AlwaysOn/Foundational-Online/_build/results?buildId=401&view=results

sebader:
before merging, should we fix the deprecation warning in AKS terraform? @heoelri

msimecek:
Since it's not a component update I suggest we create a separate work item for it.

heoelri:
> Since it's not a component update I suggest we create a separate work item for it.

I need a bit - it's already in e2e for validation.

heoelri (approved):

PR #147, opened 2022-02-23
lock state store

heoelri:


heoelri:
This is only applied at storage account / resource group creation time - this does not make too much sense. Will change it to set the locks at every execution.

heoelri:
Testing it here: https://dev.azure.com/AlwaysOn/Foundational-Online/_build/results?buildId=402&view=logs&s=324f4ed7-bdf4-5a39-9186-37c4a0e7ec24&j=9a62b2c6-931b-5ee5-4062-7dadbfc43efc

sebader:
why do we also need a lock at the RG? you wont be able to delete the RG anyway if there is a resource in there with a lock, will you?

heoelri:
True, its a bit redundant. Will drop the resource group.

heoelri:
image

Looks good.

heoelri:
Sure thing.

PR #148, opened 2022-02-23
Update online architecture diagram

hjscherer:
Change to SVG format and providing labels for architecture components


sebader:
Do we explain anywhere what the letters A-D mean? :)

hjscherer:
> Do we explain anywhere what the letters A-D mean? :)

will add it

hjscherer:
Will be updated later

PR #186, opened 2022-03-02
Component CI builds

msimecek:
As the first step to address the testing of "non-app" code components (like load generator, grafana panel etc.), mostly for reliable dependency updates, I'm kicking off a set of GitHub workflows which just do a simple restore & build of these components.

The idea is to have quick checks, which **don't require deployment**, here on GitHub and trigger them for PRs, while deployments and e2e testing will remain in ADO.

Ultimately we should figure out meaningful tests for these components, to validate that updates didn't break them.

Test runs here: https://github.com/msimecek/AlwaysOn-Foundational-Online/actions


.github/workflows/CI-Grafana.yml
```diff
@@ -0,0 +1,33 @@
...
+    working-directory: "src/infra/monitoring/grafana/healthmodelpanel"
+
+jobs:
+
+  build-and-test:
+    name: Build health model panel
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
```
heoelri, line 23:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/CI-LoadGenerator.yml
```diff
@@ -0,0 +1,52 @@
...
+  run:
+    working-directory: "src/testing/userload-generator/AzureFunctions"
+
+jobs:
+  build-and-test-global:
+    name: Build global orchestrator
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
```
heoelri, line 23:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/CI-LoadGenerator.yml
```diff
@@ -0,0 +1,52 @@
...
+          cd GlobalOrchestrator
+          dotnet build
+
+
+  build-and-test-regional:
+    name: Build regional load generator
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
```
heoelri, line 41:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/CI-SloProcessor.yml
```diff
@@ -0,0 +1,31 @@
...
+  run:
+    working-directory: "src/infra/monitoring/AlwaysOn.SloProcessor"
+
+jobs:
+  build-and-test:
+    name: Build SloProcessor
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
```
heoelri, line 22:
```suggestion
      - uses: actions/checkout@v3
```

heoelri:
As discussed earlier - i like the approach of having some pre-checks, maybe even potentially blocking gates for each PR to `main` and perhaps also to `component-updates` to do some initial upfront testing. Things that can be tested before e2e and in some cases things that are not, or cannot tested in e2e.

heoelri:
Reg. the mix of Azure Pipelines and GitHub Actions, as of now ADO/Azure Pipelines does only contain deployment related pipelines. Which makes sense to have everything consolidated in a single place/tool. In GitHub we've already some more repository-related actions like for example dependabot and perhaps more in the future. So I'm ok with adding testing/development related tasks in GitHub Actions.

heoelri:
@nielsams / @hjscherer / @CalCof widening the audience a bit. 

I think it's a good idea to have some automated testing in place as part of each pull request to `main` and `component-updates` to do some initial testing. It will of course not test 100% but it gives us at least an indicator if an update introduces a (build) breaking change.

heoelri:
Let's discuss it in todays core team meeting and perhaps pilot it in the online repo for a while then.

heoelri (approved):

PR #192, opened 2022-03-08
Improve http logging

sebader:
- dont log traces for stamp/liveness probes
- log user agent (so we can filter, e.g. AFD health probes)


msimecek (approved):
Nicer telemetry and less of it, great 👍

PR #194, opened 2022-03-09
Add Owner permission in pre-reqs

sebader:
Make this even more clear

### DIFF
--- a/docs/reference-implementation/Getting-Started.md
+++ b/docs/reference-implementation/Getting-Started.md
@@ -17,7 +17,7 @@ This guide describes the end-to-end process for setting up all pre-requisites an
 The Azure Mission-Critical reference implementation gets deployed into an Azure Subscription. For this you need to have either:
 
 - **Owner permissions** yourself and to be able to create a Service Principal which you are able to grant Owner as well, or
-- A pre-provided Service Principal with Owner permissions on a subscription
+- A pre-provisioned Service Principal with Owner permissions on a subscription
 
 The following must be installed on the client machine used to deploy Azure Mission-Critical reference implementation:
 
--- a/docs/reference-implementation/Getting-Started.md
+++ b/docs/reference-implementation/Getting-Started.md
@@ -16,7 +16,7 @@ This guide describes the end-to-end process for setting up all pre-requisites an
 
 The Azure Mission-Critical reference implementation gets deployed into an Azure Subscription. For this you will need a **Service Principal (SPN) with Owner permissions on that subscription.**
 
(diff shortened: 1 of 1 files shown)


docs/reference-implementation/Getting-Started.md
```diff
@@ -14,6 +14,11 @@ This guide describes the end-to-end process for setting up all pre-requisites an
 
 ## Pre-requisites
 
+The Azure Mission-Critical reference implementation gets deployed into an Azure Subscription. For this you need to have either:
+
+- **Owner permissions** yourself and to be able to create a Service Principal which you are able to grant Owner as well, or
+- A pre-provided Service Principal with Owner permissions on a subscription
```
heoelri, line 20:
```suggestion
- A pre-provisioned Service Principal with Owner permissions on a subscription
```

LGTM - but isn't pre-provisioned the more common terminology?

docs/reference-implementation/Getting-Started.md
```diff
@@ -14,6 +14,11 @@ This guide describes the end-to-end process for setting up all pre-requisites an
 
 ## Pre-requisites
 
+The Azure Mission-Critical reference implementation gets deployed into an Azure Subscription. For this you need to have either:
+
+- **Owner permissions** yourself and to be able to create a Service Principal which you are able to grant Owner as well, or
```
hjscherer, line 19: Owner permission doesn't allow the user to create an SPN. Can we separate it and we can also mention that User Access Administrator (UAA) would also be enough to make a role assignment for the SPN

heoelri (approved):

hjscherer (changes requested):

docs/reference-implementation/Getting-Started.md
sebader: hmm good point. What permissions do you need to create an SPN?

heoelri: Aren't these the two viable options we have here:

a) use a pre-provisioned service principals with subscription-level owner permissions 
b) auto-create a spn as part of the creation of a service connection in azure devops (what permissions does that require)?

Option a) implies that you have to get one (or create one yourself) 

We're not global administrators in the microsoft tenant, but we can use the auto create option in ADO, can't we? And we can also manually create SPs.

hjscherer: App registrations in the User Settings on the tenant
image

sebader: how about now?

hjscherer: added a minor change

hjscherer (approved):

PR #225, opened 2022-03-14
add default api version to swagger

sebader:
image

### DIFF
--- a/src/app/AlwaysOn.CatalogService/CatalogServiceHelpers.cs
+++ b/src/app/AlwaysOn.CatalogService/CatalogServiceHelpers.cs
@@ -1,22 +1,21 @@
 ﻿using AlwaysOn.Shared;
 using AlwaysOn.Shared.Exceptions;
 using AlwaysOn.Shared.Interfaces;
-using AlwaysOn.Shared.Models;
 using AlwaysOn.Shared.Models.DataTransfer;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
 using System.Net;
-using System.Text.Json;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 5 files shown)


src/app/AlwaysOn.CatalogService/CatalogServiceHelpers.cs
```diff
@@ -1,22 +1,21 @@
...
 using System.Net;
-using System.Text.Json;
 using System.Threading.Tasks;
 
 namespace AlwaysOn.CatalogService
 {
     public static class CatalogServiceHelpers
     {
+        public static int DefaultApiVersionMajor = 1;
+        public static int DefaultApiVersionMinor = 0;
```
msimecek, line 17: Are you going to use this in the other places where version is referenced?

https://github.com/Azure/Mission-Critical-Online/blob/595748184a6c438ffe5482c326efe45b5598f3dc/src/app/AlwaysOn.CatalogService/Startup.cs#L78

https://github.com/Azure/Mission-Critical-Online/blob/595748184a6c438ffe5482c326efe45b5598f3dc/src/app/AlwaysOn.CatalogService/Startup.cs#L58

msimecek (changes requested):

src/app/AlwaysOn.CatalogService/CatalogServiceHelpers.cs
sebader: yeah I was wondering: is the API definition version the same?

msimecek: For the API definition that's more of a convenience thing - not necessary. But you hese new constants can be used for `o.DefaultApiVersion`, right?

sebader: oh gotcha! I actually missed to port that part. Already did that in -Connected. fixed

msimecek (approved):

PR #229, opened 2022-03-17
Troubleshooting docs on AFD deletion error

sebader:


docs/reference-implementation/Troubleshooting.md
```diff
@@ -50,6 +50,29 @@ Sorry, we are currently experiencing high demand in this region, and cannot fulf
...
+**Solution:** You can disable the protection by running the following command towards your Azure subscription:
+
+```powershell
+az feature register --namespace Microsoft.Network --name BypassCnameCheckForCustomDomainDeletion
+
+# Then you can check the state of it by running:
+az feature list -o table --query "[?contains(name, 'Microsoft.Network/BypassCnameCheckForCustomDomainDeletion')].{Name:name,State:properties.state}
+```
+
+Once the feature has been registered (this can take a couple of minutes), deletion should work fine.
```
msimecek, line 74: Can you also add steps to "unregister" it for completeness?

docs/reference-implementation/Troubleshooting.md
```diff
@@ -50,6 +50,29 @@ Sorry, we are currently experiencing high demand in this region, and cannot fulf
...
+**Error:**
+
+```console
+Error: deleting Front Door (Subscription: "xxxxx-8cbd-46f2-a146-yyyyyyyyyy"
+│ Resource Group Name: "xxxxx-global-rg"
+│ Front Door Name: "xxxxx-global-fd"): performing Delete: frontdoors.FrontDoorsClient#Delete: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="Conflict" Message="Cannot delete frontend endpoint \"xxxxx.e2e.example.com\" because it is still directly or indirectly (using \"afdverify\" prefix) CNAMEd to front door \"xxxxx-global-fd.azurefd.net\". Please remove the DNS CNAME records and try again."
+
+```
+
+**Description:** *This only happens when you are using custom domain names*. In order to protect customers from DNS-rebinding attacks, by default an Azure Front Door resource cannot be deleted while still a CNAME is pointing to it. However, because of the dependency tree of Terraform, there is no direct way to circumvent this.
```
msimecek, line 63: > However, because of the dependency tree of Terraform, there is no direct way to circumvent this.

I don't quite understand this sentence. How is Terraform dependency tree related to the domain check on the FD resource? Shouldn't it be something about TF not able to remove the CNAME binding?

msimecek (changes requested):

docs/reference-implementation/Troubleshooting.md
msimecek: Isn't that permanent? You register the feature to disable protection and then it stays that way?

sebader: yes it is permanent. But I'm pretty sure if you ever run into this, chances are you would hit it all the time. thus you dont want to deregister. Plus, the deregister command is pretty self-explaning ;)

docs/reference-implementation/Troubleshooting.md
sebader: yeah its the shortened version of whats going on... I wasnt sure if I wanted to go into backwards-/forwards dependencies with create and destroy?!

msimecek: The sentence right now sounds like the TF dependency tree is the reason why you can't circumvent the protection, but that's not accurate, is it?

"However, because Terraform is not able to remove the CNAME dependency, it fails with this error." ?

hjscherer (approved):

docs/reference-implementation/Troubleshooting.md
sebader: sorry, somehow got lost :D now better?

docs/reference-implementation/Troubleshooting.md
sebader: also slightly reworded

msimecek (approved):

PR #231, opened 2022-03-21
Adding event hub health model

nielsams:


nielsams:
Closing because branch change

PR #238, opened 2022-03-22
Transition to use AppInsights connection string

sebader:
Not yet using it for Serilog as it breaks correlation


msimecek:
Plus the option to disable adaptive sampling in the .NET SDK for a particular environment to capture all telemetry and full dependencies.

```
ai_adaptive_sampling          = false
```

image

msimecek:
Our current implementation of IOptions doesn't support config reload after start:

image

Looking into IOptionsMonitor instead.

msimecek (approved):
The best collaborative PR in a long time! Well done ⭐

PR #239, opened 2022-03-22
Round average rating to 1 decimal number

msimecek:
Before:

image

After:

image

https://afe2e8334-global-fd.azurefd.net/#/catalog/6892b16b-91d7-4a1d-bb4a-c8f22f0efe20


msimecek:
I optimized the calculation ;) ;) It's less accurate, but FASTER!

sebader:
I realize, wouldnt it be better to round on the API side?

msimecek:
Yeah, it could... I just decided that this is more of a presentation thing, so the UI gets raw data and handles it on it's own.

sebader (approved):

PR #246, opened 2022-03-23
Removing private endpoints from Online RI

sebader:
For now this is a proposal to remove the private endpoints from the -Online RI. We still of course have them in the -Connected RI.

Reasons:
- Faster deployment
- Cleaner resource groups (no DNS zones, PE, NICs)
- Cleaner Terraform templates
- Easier debugging since you can directly connect e.g. to CosmosDB without IP whitelisting

So overall: just easier learning curve again. 
Downside: Does not appear to be as secure as before. But then again, we were actually not enforcing private connections-only to Key Vault (only to Cosmos and Event Hub). So I would argue the easier learning outweighs this.

Thoughts? @hjscherer @heoelri @nielsams @msimecek


msimecek:
> * Faster deployment

😍😍

heoelri:
The only/main reason why this was implemented in the first place is that this was brought up at first in every single review/conversation/.. am open to drop it out of the online version. Or to keep it. Don't have a strong preference here.

sebader:
> The only/main reason why this was implemented in the first place is that this was brought up at first in every single review/conversation/.. am open to drop it out of the online version. Or to keep it. Don't have a strong preference here.

yep. For anybody security-conscious, we would point them to -Connected now anyway

heoelri:
I think we'd Threat Modeling in our backlog - maybe we should do that exercise first and decide then.

sebader:
I'd also say to do that only on -Connected. Even with the Private Endpoints here we are plenty of gaps as we leave Key Vault etc open to the public

sebader:
Talked to @CalCof and he also agreed it's ok to remove the PEs from -online

@heoelri would you mind giving this a review? :) thanks!

heoelri:
> Talked to @CalCof Callum Coffin FTE and he also agreed it's ok to remove the PEs from -online
> 
> @heoelri Heyko Oelrichs FTE would you mind giving this a review? :) thanks!

Yes, for sure. When not using PE - does it make sense to consider using vNet-integration and per-Service firewall settings to lock down access to specific services then? First example that comes to my mind are the Storage Accounts used. Key Vault needs to be accessed from the hosted agents, ACR as well. And afaik is Azure DevOps (Microsoft-hosted Agents) not in the list of "Trusted Service"(s).

sebader:
The only services for which we could do that are Event Hub and the private storage account. For the others we would need special handling for the build agents and for cosmos it's not even reasonably possible with service endpoints. 
Hence I would opt to not use network restrictions at all here

heoelri:
> The only services for which we could do that are Event Hub and the private storage account. For the others we would need special handling for the build agents and for cosmos it's not even reasonably possible with service endpoints. Hence I would opt to not use network restrictions at all here

Yeah, and even for the rest is the benefit is fairly limited - it prevents access from unknown/unallowed sources, but it doesn't prevent data exfiltration etc.

src/infra/workload/releaseunit/modules/stamp/network.tf
```diff
@@ -66,7 +62,11 @@ resource "azurerm_subnet" "kubernetes" {
...
   service_endpoints = [
-    "Microsoft.Storage"
+    "Microsoft.Storage",
+    "Microsoft.AzureCosmosDB",
+    "Microsoft.KeyVault",
+    "Microsoft.ContainerRegistry",
+    "Microsoft.EventHub"
   ]
 
   enforce_private_link_endpoint_network_policies = true
```
heoelri, line 72: Based on the docs I'd assume we don't need to set this anymore.

> [enforce_private_link_endpoint_network_policies](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/subnet#enforce_private_link_endpoint_network_policies) - (Optional) Enable or Disable network policies for the private link endpoint on the subnet. Setting this to true will Disable the policy and setting this to false will Enable the policy. Default value is false.

I'd like to drop it and give it a shot.

heoelri (approved):
LGTM

PR #248, opened 2022-03-24
Add EventHub to the health model

nielsams:


nielsams:
Took a bit longer than expected. This was because I had to figure out how to deal with eventhuh having significantly higher metrics ingestion latency than other services. (5-8 min vs 2-3). This means that for a given time, min(eventhub,other source) may return healthy even though eventhub is not, because no data is defaulted to 0 in make-series. I have now done the following:
- For metrics such as processing backlog, instead of defaulting to 0 we fill forward (i.e. hold the last value until overridden).
- Consistenly done make-series instead of summarize, up to now-2m. 
- Combined summarize and make-series from different queries, where possible
- Used minimum value of TimeGenerated for a given dataset instead of hardcoded -2d. This gives greater flexibility when displaying history and when the time limit is properly implemented at the top-level query, perf isn't an issue.

sebader:
sounds great! tbh, hard to review these queries - but I trust you it works ;) Is it deployed already somewhere? And can we maybe update the screenshots of the panel that we have somewhere in the docs as well?

nielsams:
> sounds great! tbh, hard to review these queries - but I trust you it works ;) Is it deployed already somewhere? And can we maybe update the screenshots of the panel that we have somewhere in the docs as well?

It's on afe2ec079. And the updated dashboard is on Grafana INT, called Solution Health DEV.

sebader (approved):
beautiful! :)

PR #253, opened 2022-03-28
Add polling file watcher

msimecek:
Although the AKS CSI driver did a good job refreshing secrets from Key Vault, our .NET applications were not reloading it automatically. This PR fixes that by adding `DOTNET_USE_POLLING_FILE_WATCHER` environmental variable.

> Some file systems, such as Docker containers and network shares, may not reliably send change notifications. Set the DOTNET_USE_POLLING_FILE_WATCHER environment variable to 1 or true to poll the file system for changes every four seconds (not configurable).

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/file-providers?view=aspnetcore-6.0#watch-for-changes

### DIFF
--- a/src/app/charts/backgroundprocessor/templates/deployment.yaml
+++ b/src/app/charts/backgroundprocessor/templates/deployment.yaml
@@ -51,7 +51,7 @@ spec:
               value: "30"
             - name: DOCKER_HEALTHCHECK_FILEPATH
               value: "/tmp/healthy"
-            - name: "DOTNET_USE_POLLING_FILE_WATCHER" # Docker is not sending file change notification reliably, this enables file system polling every 4 seconds (not configurable) and thus automatic configuration refresh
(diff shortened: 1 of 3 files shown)


sebader:
how about explaining in the comment why we need this in the first place (because we are using KeyVault CSI driver with auto refresh)

msimecek:
I was hoping that this would do it :D

> and thus automatic configuration refresh

Will add more.

msimecek:
How about now?

sebader (approved):
Love it

heoelri (approved):
Cool!

sebader:
and to -connected, too, please :)

msimecek:
Validated in `afe2e3599`. Change of `Global-StorageAccount-ConnectionString` in Key Vault got picked up by the API in a few minutes.

PR #255, opened 2022-03-29
scale down INT AKS nodes

sebader:


msimecek (approved):

heoelri:
Can all our pods run on a single node?

heoelri:
> Can all our pods run on a single node?

Asking cause otherwise it will scale out to 2 or three anways.

sebader:
> Can all our pods run on a single node? 

Actually e2e is also set to one but I believe it always runs on 2 nodes :)

sebader:
But yeah good point, maybe we should just set both to 2 to start with

heoelri:
I assume the reason is the current quota limitation, right? The new request form does still not work and the support ticket wasn't processed yet.

image

sebader:
> I assume the reason is the current quota limitation, right? The new request form does still not work and the support ticket wasn't processed yet.
> 
> image

yeah, I also tried that :( 

what do you think, should we set the minimum for both e2e and int to 2nodes since that seems to be what we require anyway?

PR #257, opened 2022-03-31
Add PodDisruptionBudgets

heoelri:
This PR adds PodDistruptionBudgets to our `CatalogService` and `HealthService` helm charts. This is based on a conversation with @hjscherer and @sebader in Teams.

We've already certain measures in place to avoid unplanned disruptions like a min. number of replicas per deployment, a horizontal pod autoscaler (hpa) configuration as well as the default deployment rollout strategy "RollingUpdate". This is an addition to reduce the impact of voluntary disruptions even further.

See [Plan for availability using pod disruption budgets](https://docs.microsoft.com/en-us/azure/aks/operator-best-practices-scheduler#plan-for-availability-using-pod-disruption-budgets), [Voluntary and involuntary disruptions](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/) and [What Is PodDisruptionBudget, And Why You Should Use It in Kubernetes](https://innablr.com.au/blog/what-is-pod-disruption-budget-in-k8s-and-why-to-use-it/) for more.


heoelri:
```console
kubectl get pdb -n workload

NAME                 MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
catalogservice-pdb   50%             N/A               1                     10m
healthservice-pdb    50%             N/A               1                     11m
```

```console
kubectl get deploy -n workload

NAME                         READY   UP-TO-DATE   AVAILABLE   AGE
backgroundprocessor-deploy   2/2     2            2           11m
catalogservice-deploy        3/3     3            3           10m
healthservice-deploy         3/3     3            3           12m
```

sebader:
makes a lot of sense to me - after I read the external docs ;) We should include at least pointer to these in our docs. And describe briefly why we add them: "to avoid downtime". I like this statement from one of the links:

> For the sake of brevity, let’s consider you have a deployment of nginx with replicas = 2. If both of them are scheduled onto a single node, what will happen if that particular node is recycled? Downtime, until the pod is re-scheduled on another node.

heoelri:
> makes a lot of sense to me - after I read the external docs ;) We should include at least pointer to these in our docs. And describe briefly why we add them: "to avoid downtime". I like this statement from one of the links:
> 
> > For the sake of brevity, let’s consider you have a deployment of nginx with replicas = 2. If both of them are scheduled onto a single node, what will happen if that particular node is recycled? Downtime, until the pod is re-scheduled on another node.

Yes I agree, though this shouldn't happen often in our case. We've always at least two nodes, in different zones, with always at least 2 replicas. K8s distributes replicas automatically across different nodes. But there might be some rare edge cases where e.g. the two healthservice replicas are running on the same node in a two node cluster setup after a cluster upgrade, pdb should help solving that.

src/config/ingress/values.helm.yaml
```diff
@@ -1,12 +1,14 @@
...
 controller:
+  securityContext:
+    readOnlyRootFilesystem: true # If a container does not need to write files, it should be run with a read-only filesystem. 
+    privileged: false # Running a container as privileged gives all capabilities to the container, and it also lifts all the limitations enforced by the device cgroup controller.
   autoscaling:
-    enabled:     true
-    minReplicas: 3
-    maxReplicas: 12 
+    enabled:     true # Enable autoscaling via horizontal pod autoscaler
+    minReplicas: 3 # Minimum number of nodes
```
sebader, line 10: cant we add PDB to ngnix as well?

heoelri: the current helm chart doesn't allow that - might be worth a contribution - let me give it a try

heoelri: it's already there - by default - the no. of `minAvailable` replicas can be configured..

```console
kubectl get pdb -n kube-system
NAME                       MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
coredns-pdb                1               N/A               1                     9d
ingress-nginx-controller   1               N/A               2                     9d
konnectivity-agent         1               N/A               1                     5d6h
```

sebader: ah cool. Can you just mentioned that in the docs as well with a sentence?

docs/reference-implementation/AppDesign-Application-Design.md
```diff
@@ -41,16 +41,20 @@ There is no backchannel which communicates to the client if the operation comple
...
 
 The `BackgroundProcessor` service has very different requirements and is considered a background worker which has no direct impact on the user experience. As such, `BackgroundProcessor` has a different auto-scaling configuration than `CatalogService` and it can scale between 2 and 32 instances (which matches the max. no. of EventHub partitions). The ratio between `CatalogService` and `BackgroundProcessor` is around 20:2.
 
+All workload components as well as supporting services like the `HealthService` and dependencies like `ingress-nginx` are configured with at least 3 or in case of the `HealthService` 2 instances (replicas) per cluster. This is supposed to prevent certain availability issues and to ensure that the service is always available. The instances are automatically spread across nodes and therefore also across Availability Zones.
+
+In addition to that does each component of the workload including dependencies like `ingress-nginx` have [Pod Disruption Budgets (PDBs)](https://docs.microsoft.com/azure/aks/operator-best-practices-scheduler#plan-for-availability-using-pod-disruption-budgets) configured to ensure that a minimum number of instances is always available.
```
sebader, line 56:
```suggestion
In addition to that, each component of the workload including dependencies like `ingress-nginx` has [Pod Disruption Budgets (PDBs)](https://docs.microsoft.com/azure/aks/operator-best-practices-scheduler#plan-for-availability-using-pod-disruption-budgets) configured to ensure that a minimum number of instances is always available.
```

src/app/charts/catalogservice/templates/pdb.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ .Chart.Name }}-pdb
+spec:
+  minAvailable: 50%
```
sebader, line 6: why here 50% and for the others absolute numbers?

heoelri: For the healthservice we've only 2 replicas to compensate a single node outage as the expected load is low. Therefore would a single replica in voluntary disruption scenarios be sufficient. 

For catalogservice a failback from let's say 10 (scaled out automatically) nodes to 1 would cause issues, 50% would still result in a significant performance degradation, but the service would remain to work, and 50% is just a starting point to balance between overhead and availability.

For backgroundprocessor we, from my pov, don't need a pdb at all given it has no user impact. But a pdb with minavailable 1 would result in at least 1 remaining replica to process data.

sebader: ok makes sense

sebader (approved):
LGTM thanks!

PR #258, opened 2022-03-31
Adding cert-manager guidance to TSG

heoelri:
This PR contains some additional guidance on how to fix cert-manager.

### DIFF
--- a/docs/reference-implementation/Troubleshooting.md
+++ b/docs/reference-implementation/Troubleshooting.md
@@ -116,7 +116,7 @@ To manually force cert-manager to re-request the certificate you can delete the 
 # Check certificate status
 kubectl get certificates -n workload
 
-# Delete certificate (in FALSE state)
+# Delete certificate (if in `ready=FALSE` state). This will trigger cert-manager to create a new certificate request
 kubectl delete certificate/<certificatename> -n workload
 ```
 
--- a/docs/reference-implementation/Troubleshooting.md
+++ b/docs/reference-implementation/Troubleshooting.md
@@ -93,9 +93,9 @@ Deployment not ready. Retrying... 5/5
 ##[error]Script failed with exit code: 1
 ```
 
-**Description:** The "Install workload CatalogService on AKS clusters" step actively monitors the provisioning of a LetsEncrypt certificate for the ingress. If the certificate is not ready, or cannot successfully provisioned this task will fail.
+**Description:** The "Install workload CatalogService on AKS clusters" step actively monitors the provisioning of a LetsEncrypt certificate for the ingress during the installation of the CatalogService helm chart. If the certificate is not ready, or cannot successfully provisioned this task will fail.
 
-**Solution:** This is most of the time caused by hitting LetsEncrypt thresholds (see [Rate Limits](https://letsencrypt.org/docs/rate-limits/) for more details) for certificate provisioning. You can try to re-run the step, or wait for the certificate to be ready.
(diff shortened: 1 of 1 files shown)


docs/reference-implementation/Troubleshooting.md
```diff
@@ -81,6 +81,35 @@ Deployment of service [HealthService | BackgroundProcessor | CatalogService] fai
...
+**Description:** The "Install workload CatalogService on AKS clusters" step actively monitors the provisioning of a LetsEncrypt certificate for the ingress. If the certificate is not ready, or cannot successfully provisioned this task will fail.
+
+**Solution:** This is most of the time caused by hitting LetsEncrypt thresholds (see [Rate Limits](https://letsencrypt.org/docs/rate-limits/) for more details) for certificate provisioning. You can try to re-run the step, or wait for the certificate to be ready.
+
+```console
+# Check certificate status
+kubectl get certificates -n workload
+
+# Check certificate request status
+kubectl get certificaterequests -n workload
```
sebader, line 105: how about adding a "k delete certificaterequest" as a possible solution? this should immediately trigger reissue

heoelri: I was thinking about that as well - what's important though is to make sure that the reason is clear - in case you're hitting a rate limit on the Let's Encrypt side you can delete it as often as you want and nothing will change.

And I'd suggest to delete the certificate not the certificaterequest. Let me rephrase it a bit.

docs/reference-implementation/Troubleshooting.md
```diff
@@ -81,6 +81,45 @@ Deployment of service [HealthService | BackgroundProcessor | CatalogService] fai
...
+
+Relevant log events are for example `cert-manager/certificates-trigger "msg"="Not re-issuing certificate as an attempt has been made in the last hour" "key"="workload/catalogservice-ingress-secret" "retry_delay"=3599055978096`.
+
+To manually force cert-manager to re-request the certificate you can delete the existing one:
+
+```console
+# Check certificate status
+kubectl get certificates -n workload
+
+# Delete certificate (in FALSE state)
```
sebader, line 119:
```suggestion
# Delete certificate (if in `ready=FALSE` state). This will trigger cert-manager to create a new certificate request
```

docs/reference-implementation/Troubleshooting.md
```diff
@@ -81,6 +81,45 @@ Deployment of service [HealthService | BackgroundProcessor | CatalogService] fai
...
+```
+
+Relevant log events are for example `cert-manager/certificates-trigger "msg"="Not re-issuing certificate as an attempt has been made in the last hour" "key"="workload/catalogservice-ingress-secret" "retry_delay"=3599055978096`.
+
+To manually force cert-manager to re-request the certificate you can delete the existing one:
+
+```console
+# Check certificate status
+kubectl get certificates -n workload
+
```
sebader, line 118: I guess we dont have a sample console output right now, that we could show?! maybe lets keep that in mind and add once we run into it again

sebader (changes requested):

heoelri:
Before this comes up - I start questioning if it makes sense to maintain the same/the overlapping TSG content (which is currently most of it) in all RIs.

sebader:
> Before this comes up - I start questioning if it makes sense to maintain the same/the overlapping TSG content (which is currently most of it) in all RIs.

yeah valid point. Since -Connected mostly adds stuff, we could keep the common issues in -Online and have in -Connected a pointer to this and then only maintain issues that are specific to Connected

docs/reference-implementation/Troubleshooting.md
heoelri: Right - I've already manually fixed and destroyed the affected environment.

sebader (approved):

sebader:
thanks for this!

PR #259, opened 2022-03-31
Allow deletion of non-empty per-stamp resource groups

heoelri:
This PR enables the deletion of non-empty resource groups. This is required as sometimes stale resources (like NICs for PEs) remain in resource groups during deletion.


sebader (approved):
Please in Connected, too. (There we actually need it)

sebader (approved):
Please in Connected, too. (There we actually need it)

PR #261, opened 2022-04-01
Change always-on.app to mission-critical.app

nielsams:


sebader (approved):
guess we will only see if it works when we deploy INT :)

nielsams:
I'll wait 24 hours before merging it in case the nameserver settings still need to propagate.

PR #276, opened 2022-04-04
Update arch-diagram 

hjscherer:
Remove PE on the architecture diagram for the online RI


sebader (changes requested):
the resolution gets lower every time. could you increase it a bit?

hjscherer:
> the resolution gets lower every time. could you increase it a bit?

Will move to SVG for better resolution. Did that already a while ago on the connected diagram

sebader (approved):
thanks for this!

heoelri (approved):
LGTM

heoelri:
> What is the status of this PR? Is this still relevant? @hjscherer Hansjoerg Scherer FTE / @sebader Sebastian Bader FTE

Ah, I see. Depends on the CODEOWNERS update.

CalCof (approved):

PR #279, opened 2022-04-05
Task retries and ingress timeout

sebader:
Learning from today's incident:
- Ingress ingress timout to higher than default of 60seconds in order to see actual errors from the backend instead of 504 Gateway timeouts
- Retry most of our ADO tasks at least once. Doesnt hurt and might "solve" many transient issues

### DIFF
--- a/.ado/pipelines/templates/jobs-workload-deploy.yaml
+++ b/.ado/pipelines/templates/jobs-workload-deploy.yaml
@@ -13,6 +13,7 @@ jobs: # using jobs so they each run in parallel
 
     - task: AzureCLI@2
       displayName: 'Install workload healthservice on AKS clusters'
+      retryCountOnTaskFailure: 2
       inputs:
         azureSubscription: $(azureServiceConnection)
         scriptType: pscore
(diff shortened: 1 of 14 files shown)


.ado/pipelines/templates/jobs-workload-deploy.yaml
```diff
@@ -160,6 +162,7 @@ jobs: # using jobs so they each run in parallel
 
     - task: AzureCLI@2
       displayName: 'Install workload BackgroundProcessor on AKS clusters'
```
heoelri, line 164: not sure if you want this set in all tasks for consistency - we've already inline retry logic for the deployment of catalogservice and backgroundprocessor (due to the issues with cert-manager)

sebader: hm valid point. would would you prefer?

heoelri: it probably makes sense to set it consistenly across all tasks as you've suggested - inline is still required when a delay between two retries is required

sebader: ok lets keep it as is then

heoelri (approved):

PR #286, opened 2022-04-05
Update package versions to be semver compatible

msimecek:
This should fix this Dependabot error in AlwaysOn.UI package.json

image


sebader (approved):
thanks!

PR #289, opened 2022-04-07
Move cert-manager install to helm and update to 1.8.0

sebader:
I realized we did not have a way to replace our (AlwaysOn) email address in there - which is actually quite important do to, otherwise everybody would use ours.
Plus, I'm using the same helm-based approach also when switching to DNS issuer which needs more inputs

https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=939&view=results


heoelri:
We haven't done it in the first place cause of the overhead for a single file, but yes, I agree from a consistency point of view it might be a good idea to wrap the issuer into a helm chart as well.

.ado/pipelines/templates/jobs-configuration.yaml
```diff
@@ -106,16 +106,23 @@ jobs:
...
-          echo "*** Create a separate namespace for cert-manager on $aksClusterName"
-          kubectl apply -f src/config/cert-manager/cert-manager.yaml
+          $certManagerNamespace = "cert-manager"
+
+          echo "*** Create namespace '$certManagerNamespace' on AKS cluster $aksClusterName via kubectl"
+          kubectl create namespace $certManagerNamespace `
+                    --dry-run=client --output yaml | kubectl apply -f -
+
+          echo "*** Setting up cert-manager ClusterIssuer with Let's Encrypt"
+          helm upgrade --install cert-manager-alwayson src/config/charts/cert-manager `
```
heoelri, line 116: would it make sense to call this chart `cert-manager-config` instead of `cert-manager-alwayson`?

heoelri (approved):
LGTM! 👍

PR #290, opened 2022-04-08
Update Grafana to use AzureRM 3.1.0 and enable zone balancing

heoelri:
This PR bumps the azurerm provider version used to deploy Grafana from `3.0.2` to `3.1.0`: https://github.com/hashicorp/terraform-provider-azurerm/releases/tag/v3.1.0

And enables zone balancing via `zone_balancing_enabled = true` for `azurerm_service_plan`.

The used `azurerm_postgresql_server` does not support zones, yet. It might be worth checking if a switch to `azurerm_postgresql_flexible_server` makes sense.

E2E currently fails when destroying (configuring `zone_balancing_enabling` forces a replacement) the `azurerm_service_plan`: https://dev.azure.com/mission-critical/Online/_build/results?buildId=951&view=logs&j=500d37e8-38c4-5dfd-d6ad-284cea37a878&t=7349e4b2-4685-544a-1082-66a60321d9d1


src/infra/monitoring/grafana/terraform/stamps/app-service.tf
```diff
@@ -6,6 +6,8 @@ resource "azurerm_service_plan" "asp" {
   os_type             = "Linux"
   sku_name            = "P1v2"
 
+  zone_balancing_enabled = true # Balance Service Plan across Availability Zones (AZs).
```
sebader, line 9: this requires at least three app service instances, I believe. Not sure if thats worth it for grafana right now. maybe just add it and comment it out (with a comment why)?

sebader (changes requested):

src/infra/monitoring/grafana/terraform/stamps/app-service.tf
heoelri: Yep, was thinking the same - especially as the current PostgreSQL instance doesn't support zones at all.

heoelri:
```suggestion
  zone_balancing_enabled = false # Balance Service Plan across Availability Zones (AZs)
  # This is currently disabled as the database backend is not using AZs.
```

sebader (approved):

PR #293, opened 2022-04-08
Draft - Import container images

heoelri:
@sebader this is how i'd envision the import of external container images e.g. for nginx-ingress and other dependencies. 

Import can be automated up to a certain degree - authentication for example is currently not part of this PR. Deployment cannot - the parameters are heavily depending on each individual helm chart and the tasks are itself also not really standardized, yet.

**This is not tested - and I'm not saying that we should do it like that.**


heoelri:
This PR now contains the images for ingress-nginx and cert-manager and imports them as part of the pipeline:

image

image

image

sebader:
Could we test this by applying a policy which will only allow images from either mcr.microsoft.com or myacr.azurecr.io?

heoelri:
> Could we test this by applying a policy which will only allow images from either mcr.microsoft.com or myacr.azurecr.io?

Yes, should be possible using something (doesn't contain the specific ACR, yet) like this: https://github.com/Azure/Mission-Critical-Online/blob/ba6a76e0b5133ad1add7aacff56a7819f8ccddf6/src/infra/workload/releaseunit/modules/stamp/policy.tf#L57-L75

src/infra/workload/releaseunit/modules/stamp/policy.tf
```diff
@@ -52,4 +52,24 @@ resource "azurerm_resource_group_policy_assignment" "aks_rbac_enabled" {
...
\ No newline at end of file
+}
+
+# Limit access to external container registries (BuiltIn)
+# https://github.com/Azure/azure-policy/blob/master/built-in-policies/policyDefinitions/Kubernetes/ContainerAllowedImages.json
+resource "azurerm_resource_group_policy_assignment" "aks_limit_registries" {
+  name                 = "aks-limit-registries"
+  resource_group_id    = azurerm_resource_group.stamp.id
+  policy_definition_id = "/providers/Microsoft.Authorization/policyDefinitions/febd0533-8e55-448f-b837-bd0e06f16469"
+  display_name         = "Kubernetes cluster containers should only use allowed images"
```
sebader, line 63:
```suggestion
  display_name         = "Kubernetes cluster containers should only be pulled from allowed container registries"
```

heoelri: The name used is the name of the policy. https://github.com/Azure/azure-policy/blob/master/built-in-policies/policyDefinitions/Kubernetes/ContainerAllowedImages.json#L3

sebader: hm fair point. But what if we find that name misleading? :D I'd assume we can call it here whatever we want?

heoelri: We of course can - and I've added a link to the GitHub repo. But, to make it easier I'd prefer to use the exact same name you'd find in the portal. Even though in this specific case it's not perfect - though its not wrong. You can use it for container image names as well as for registries.

heoelri: Maybe you'd want to raise a PR here: https://github.com/Azure/azure-policy/blob/master/built-in-policies/policyDefinitions/Kubernetes/ContainerAllowedImages.json#L3

.ado/pipelines/templates/jobs-configuration.yaml
```diff
@@ -22,6 +22,7 @@ jobs:
       inlineScript: |
 
         # load json data from downloaded pipeline artifact json
+        $globalInfraDeployOutput = Get-ChildItem $(Pipeline.Workspace)/terraformOutputGlobalInfra/*.json | Get-Content | ConvertFrom-JSON
```
heoelri, line 25: I thought I was using `steps-parse-terraform-output.yaml` for the global terraform values instead. Have you dropped it?

sebader: yes, to keep it consistent with the other places where we use the outputs. the parse function is only used/needed if we dont use the output in a script directly

heoelri: I mean we only need `acr_name` here, but wouldn't it be a bit cleaner and easier to read to run the `steps-parse-terraform-output.yaml` at the top of each job instead?

src/infra/workload/releaseunit/modules/stamp/policy.tf
```diff
@@ -52,4 +52,25 @@ resource "azurerm_resource_group_policy_assignment" "aks_rbac_enabled" {
...
+
+  parameters = <<PARAMS
+  {
+    "allowedContainerImagesRegex": {
+      "value": "^(${data.azurerm_container_registry.global.name}\\.azurecr\\.io|mcr\\.microsoft\\.com)\\/.+$"
+    },
+    "excludedNamespaces": {
+      "value": "[]"
```
sebader, line 72:
```suggestion
      "value": []
```

.ado/pipelines/templates/jobs-configuration.yaml
```diff
@@ -61,9 +64,7 @@ jobs:
                 --values src/config/ingress-nginx/values.helm.yaml `
                 --set controller.service.loadBalancerIP="$aksIngressIp" `
                 --set controller.service.annotations."service\.beta\.kubernetes\.io/azure-load-balancer-resource-group"="$aksClusterResourceGroup" `
-                --version "$(ingressNginxVersion)" `
-                --set controller.image.repository="$($globalInfraDeployOutput.acr_login_server.value)/$($container.image)" `
-                --set controller.image.tag="$($container.tag)"
+                --version "$(ingressNginxVersion)" $privateRegistryParam
```
heoelri, line 67: @sebader thanks! I prefer that approach - does it work now? Or have you just reverted it for testing purposes?

.ado/pipelines/templates/jobs-configuration.yaml
sebader: hmm good question. This would only work for the global infra output, not for the release unit. I'm can't say I'd prefer either way.
@msimecek thoughts from your side?

heoelri: True, the JSON format for the release unit is slightly different and needs to support 1 or more release units. Could be challenging to translate that into meaningful variables.

I'd only the global infra output in mind - we've `steps-parse-terraform-output.yaml`, we're already using it - so why not using it consistently accross all stages? `$(acr_name)` is definitely easier to read than `$($globalInfraDeployOutput.acr_login_server.value)` ;)

sebader: yes agree, it would be much easier to read. I'm just not sure if it will be more confusing to people since we might then have two different ways to read the variables in the same scripts?

sebader:
yeah sure, we can also still keep the branch around

PR #294, opened 2022-04-08
Draft import workload images and helm charts

sebader:


.ado/pipelines/config/workloads.json
```diff
@@ -0,0 +1,35 @@
+{
+    "catalogservice": {
+        "chart": {
+            "name": "afintglobalcr.azurecr.io/helm/catalogservice",
```
heoelri, line 4: Any specific reason to use a different schema for charts and images?

.ado/pipelines/templates/jobs-workload-deploy.yaml
```diff
@@ -20,10 +20,41 @@ jobs: # using jobs so they each run in parallel
...
+          {
+            throw "*** Workload '$workloadName' not found in $workloadsFile"
+          }
+
+          $chartRegistry,$chartName = $workload.chart.name -split "/",2
+
+          # if the helm chart is hosted on ACR, we need to log in first
+          if($chartRegistry.EndsWith(".azurecr.io"))
+          {
+            az acr login -n $chartRegistry
```
heoelri, line 53: This is something I've though about in #293 but not implemented it yet. You're assuming that service connection has access to the source registry, right?

sebader: yep, I would start with that. Otherwise it will get messy

.ado/pipelines/config/workloads.json
sebader: not really. I was just not sure if the name for helm charts will always be in the same format as for images? (registry/name)

PR #305, opened 2022-04-19
Adding Kured for node reboots

sebader:
This PR installs Kured on the AKS clusters to handle node reboots which might be required do to nightly auto-updates, e.g. when Kernel updates need to be applied.

Still testing, I have an open deployment which should required (and do) the reboot in the next days/nights

Docs: https://docs.microsoft.com/en-us/azure/aks/node-updates-kured

@heoelri

### DIFF
--- a/.ado/pipelines/config/configuration.yaml
+++ b/.ado/pipelines/config/configuration.yaml
@@ -14,6 +14,8 @@ variables:
   value: '4.0.18'
 - name: 'certManagerVersion'  # cert-manager helm chart version
   value: 'v1.8.0'
+- name:  'kuredVersion'        # KUbernetes REboot Daemon helm chart version
+  value: '2.13.0'
 - name: 'dotnetSdkVersion'    # dotnet sdk version
   value: '6.0.200'
 - name: 'chaosMeshVersion'    #  chaos-mesh chart version
--- a/.ado/pipelines/templates/jobs-configuration.yaml
+++ b/.ado/pipelines/templates/jobs-configuration.yaml
@@ -204,14 +204,16 @@ jobs:
           az aks get-credentials --name $aksClusterName --resource-group $aksClusterResourceGroup --overwrite-existing
           echo "*** Installing the KUbernetes REboot Daemon (kured) via helm on $aksClusterName"
 
-          echo "*** Create namespace 'kured' on AKS cluster $aksClusterName via kubectl"
-          kubectl create namespace kured `
+          $kuredNamespace = "kube-system"
+
+          echo "*** Create namespace '$kuredNamespace' on AKS cluster $aksClusterName via kubectl"
+          kubectl create namespace $kuredNamespace `
                     --dry-run=client --output yaml | kubectl apply -f -
 
           helm repo add kured https://weaveworks.github.io/kured
           helm repo update
           helm upgrade --install kured kured/kured `
-                        --namespace kured `
+                        --namespace $kuredNamespace `
                         --values src/config/kured/kured.yaml `
                         --version $(kuredVersion)
         }
--- a/.ado/pipelines/templates/jobs-configuration.yaml
+++ b/.ado/pipelines/templates/jobs-configuration.yaml
@@ -183,6 +183,38 @@ jobs:
                         --set kvSecrets="{$allSecretNames}"
         }
 
+  - task: AzureCLI@2
+    displayName: 'Install kured on AKS clusters'
+    inputs:
+      azureSubscription: $(azureServiceConnection)
+      scriptType: pscore
+      scriptLocation: inlineScript
+      inlineScript: |
+        # load json data from downloaded pipeline artifact json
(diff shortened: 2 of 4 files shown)


sebader:
@heoelri I tried putting this in its own namespace but that did not work as it requires privileged containers. Any hint how we could enable this on other namespaces except for kube-system? Or should we just leave it in system?

heoelri:
Am ok with adding kured back (see https://github.com/Azure/AlwaysOn-Engineering/pull/443 where we've dropped it in favor of the back then new AKS-native node-image update process) to reboot linux-based nodes with pending reboots (updates) installed via Ubuntu's (resp. Debian) by default enabled [UnattendedUpgrade](https://wiki.debian.org/UnattendedUpgrades) mechanism. The recently added pod disruption budgets (PR #257) should further limit the impact of updates/node reboots to an acceptable minimum.

Our current implementation (w/o kured) has configured a weekly maintenance window (using [planned maintenance](https://docs.microsoft.com/en-us/azure/aks/planned-maintenance) - in preview): 
https://github.com/Azure/Mission-Critical-Online/blob/69f84b008f72b224d3254c1f3a19a371c88c8dec/src/infra/workload/releaseunit/modules/stamp/kubernetes.tf#L13-L18

Where to execute [AKS auto upgrade](https://docs.microsoft.com/en-us/azure/aks/upgrade-cluster#set-auto-upgrade-channel) for node image upgrades:

https://github.com/Azure/Mission-Critical-Online/blob/69f84b008f72b224d3254c1f3a19a371c88c8dec/src/infra/workload/releaseunit/modules/stamp/kubernetes.tf#L11

The node-image upgrade process on a weekly basis in combination with weekly released AKS node images can be sufficient for most scenarios. Using kured can help to further minimize the gap between two image releases and node image upgrades. By rebooting nodes with pending reboots.

> "[..] Linux node images are updated weekly, and Windows node images updated monthly. [..]"
> https://docs.microsoft.com/en-us/azure/aks/node-image-upgrade

From my POV we've to make sure that kured isn't executed on each regional stamp at the same time to limit its operations to one region at a time maybe by using the region's local timezone. And we should further investigate, for example by running some load during kured's execution window, the impact its operations have on a stamps availability.

sebader:
> From my POV we've to make sure that kured isn't executed on each regional stamp at the same time to limit its operations to one region at a time maybe by using the region's local timezone. 

Thats a good point! I'll try to think of something. The region's timezone will probably not be good enough since in many cases customers might run their workload and regions which are in the same geo

sebader:
hmm, last night kured did....something. Not quite sure I fully understand the behavior. Let's have a look together once you are back

heoelri:
> @heoelri Heyko Oelrichs FTE I tried putting this in its own namespace but that did not work as it requires privileged containers. Any hint how we could enable this on other namespaces except for kube-system? Or should we just leave it in system?

Ok for me to put it into the `kube-system` namespace for now. The other alternative would be modifying the policy that restricts privileged pods. Which by default only excludes `kube-system` (afaik). But that's outside of this RI of course a bit more effort and complexity.

https://github.com/Azure/Mission-Critical-Online/blob/69f84b008f72b224d3254c1f3a19a371c88c8dec/src/infra/workload/releaseunit/modules/stamp/policy.tf#L39-L46

> > From my POV we've to make sure that kured isn't executed on each regional stamp at the same time to limit its operations to one region at a time maybe by using the region's local timezone.
> 
> Thats a good point! I'll try to think of something. The region's timezone will probably not be good enough since in many cases customers might run their workload and regions which are in the same geo

True, we could also think about different weekdays or so - for kured and probably also for planned maintenance where we currently also use the same timewindow which is also not great. 

> hmm, last night kured did....something. Not quite sure I fully understand the behavior. Let's have a look together once you are back

👍

heoelri:
One foundational question (the reason why we've removed it in the first place) remains, do we want and need kured at all. We've weekly planned maintenance windows scheduled and AKS configured to automatically upgrade to newer node image versions.

> Linux node images are updated weekly, and Windows node images updated monthly. Although customers will be notified of image upgrades via the AKS release notes, it might take up to a week for updates to be rolled out in all regions. 
> https://docs.microsoft.com/en-us/azure/aks/node-image-upgrade

On Linux nodes - Updates are applied automatically (even w/o kured) via Debian/Ubuntu's UnattendedUpgrade. The only thing that kured does is to reboot nodes with pending reboots. As part of every scale (-out) and upgrade operation AKS nodes will fall back to the currently used node image version w/o any updates installed. UnattendedUpgrade will then kick it in again, and kured will reboot the patched node again when needed.

> Unattended upgrades apply updates to the Linux node OS, but the image used to create nodes for your cluster remains unchanged. If a new Linux node is added to your cluster, the original image is used to create the node. This new node will receive all the security and kernel updates available during the automatic check every night but will remain unpatched until all checks and restarts are complete.
> https://docs.microsoft.com/en-us/azure/aks/node-updates-kured#node-image-upgrades

PR #309, opened 2022-04-25
Configure ingress-nginx `externalTrafficPolicy` to fix issues

heoelri:
This PR bumps the `ingress-nginx` chart from `4.0.18` to `4.1.0` and configures the `externalTrafficPoliy` to `Local`.

> Preserving the client source IP
>By default, the source IP seen in the target container is not the original source IP of the client. To enable preservation of the client IP, the following fields can be configured in the .spec of the Service:
> 
> `.spec.externalTrafficPolicy` - denotes if this Service desires to route external traffic to node-local or cluster-wide endpoints. There are two available options: Cluster (default) and Local. Cluster obscures the client source IP and may cause a second hop to another node, but should have good overall load-spreading. Local preserves the client source IP and avoids a second hop for LoadBalancer and NodePort type Services, but risks potentially imbalanced traffic spreading.
> https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/

### DIFF
--- a/.ado/pipelines/config/configuration.yaml
+++ b/.ado/pipelines/config/configuration.yaml
@@ -11,7 +11,7 @@ variables:
 - name: 'helmVersion'         # helm package manager version
   value: 'v3.8.1'
 - name: 'ingressNginxVersion' # nginx ingress controller helm chart version
-  value: '4.0.18'
+  value: '4.1.0'
 - name: 'certManagerVersion'  # cert-manager helm chart version
   value: 'v1.8.0'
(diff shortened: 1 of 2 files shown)


heoelri:
This results in a properly configured health probe.

image

https://dev.azure.com/mission-critical/Online/_build/results?buildId=1129&view=results

sebader (approved):

heoelri:
> I would opt to merge this PR now. we can still revert if this get fixed somewhere else

Ok for me - feel free to merge. 

Support case is still open and we've two envs running affected by this issue.

PR #332, opened 2022-05-04
Rename ADO displayNames for loadtesting

heoelri:
This PR renames some of the displayNames used for the embedded and standalone loadtest.


heoelri:
Running e2e now and will paste some screenshots here as soon as it's done. Will use the same displayNames in the other LT implementation as well to reduce the need to update screenshots later. @hjscherer FYI

.ado/pipelines/templates/stages-locust.yaml
```diff
@@ -22,10 +22,10 @@ parameters:
 stages:
```
sebader, line 22: I guess we leave the file name untouched for now and later might have two different stage files for ALT and Locust?

heoelri: Yes. The other one already exists. For now its just about the displayNames.

sebader (approved):

heoelri:
@hjscherer renamed:

image

PR #341, opened 2022-05-05
Pull `grafana` and `nginx` from `mcr.microsoft.com`

heoelri:
This PR updates the Dockerfiles for grafana and the AlwaysOn.UI to pull `grafana` and `nginx` from `mcr.microsoft.com` instead of `hub.docker.com`. In preparation for the announced throttling of Docker Hub:

> Docker Hub limits the number of Docker image downloads (“pulls”) based on the account type of the user pulling the image. Pull rates limits are based on individual IP address. For anonymous users, the rate limit is set to 100 pulls per 6 hours per IP address. For [authenticated](https://docs.docker.com/docker-hub/download-rate-limit/#how-do-i-authenticate-pull-requests) users, it is 200 pulls per 6 hour period.
> https://docs.docker.com/docker-hub/download-rate-limit

### DIFF
--- a/src/app/AlwaysOn.UI/Dockerfile
+++ b/src/app/AlwaysOn.UI/Dockerfile
@@ -10,7 +10,7 @@ RUN npm install && \
     npm run build
 
 # Create application container
-FROM nginx:alpine
+FROM mcr.microsoft.com/oss/nginx/nginx:1.17.5-alpine
 # Copy build artifacts from previous stage build-env
 COPY --from=build-env /app/dist /usr/share/nginx/html
 EXPOSE 80
--- a/src/infra/monitoring/grafana/Dockerfile
+++ b/src/infra/monitoring/grafana/Dockerfile
@@ -18,7 +18,7 @@ RUN npm install
 
 # We're using Grafana 8.3.4 now since later versions give us issues with the header settings behind Front Door. 
 # As soon as that is resolved, we'll get back on the latest version. 
-FROM grafana/grafana:8.3.4
+FROM mcr.microsoft.com/oss/grafana/grafana:8.3.4
 
 # In order to run unsigned plugins such as our health model panel, we need to explicitly allow them in an env variable. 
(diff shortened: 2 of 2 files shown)


heoelri:
"Successfully" tested in E2E: https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=1227&view=results

image

Discussing with the team how to get a replacement for `node`.

src/app/AlwaysOn.UI/Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN npm install && \
     npm run build
 
 # Create application container
-FROM nginx:alpine
+FROM mcr.microsoft.com/oss/nginx/nginx:1.17.5-alpine
```
sebader, line 13: so far this was using latest. is there a reason why we are going to a specific one? While :latest has its own issues, we otherwise have to keep it up to date. given that we are not using this Dockerfile really anywhere, I'd opt for latest

sebader:
side question: how did you figure out which images are available in mcr? :)

src/app/AlwaysOn.UI/Dockerfile
heoelri: Mainly cause there's no `:latest` - in theory dependabot should take care of it.

heoelri: Besides that, I don't think that we use `latest` anywhere else. 

https://github.com/Azure/Mission-Critical-Online/blob/8b21fa40b97323aa84caf6de978c27e07f4a3a6e/src/infra/monitoring/grafana/Dockerfile#L2

https://github.com/Azure/Mission-Critical-Online/blob/8b21fa40b97323aa84caf6de978c27e07f4a3a6e/src/infra/monitoring/grafana/Dockerfile#L21

https://github.com/Azure/Mission-Critical-Online/blob/8b21fa40b97323aa84caf6de978c27e07f4a3a6e/src/app/AlwaysOn.UI/Dockerfile#L2

So might be a good opportunity to align that.

heoelri:
I assume we ignore that for now and will continue to use the optout flag till it stops working?

PR #344, opened 2022-05-06
Exclude system namespaces from Azure Policy assignments

heoelri:
This PR excludes some Kubernetes system namespaces like `kube-system` and `chaos-testing` from Azure Policy assignments.


src/infra/workload/releaseunit/modules/stamp/policy.tf
```diff
@@ -25,6 +25,19 @@ resource "azurerm_resource_group_policy_assignment" "no_privilege_escalation" {
   resource_group_id    = azurerm_resource_group.stamp.id
   policy_definition_id = "/providers/Microsoft.Authorization/policyDefinitions/1c6e92c9-99f0-4e55-9cf2-0c234dc48f99"
   display_name         = "Kubernetes clusters should not allow container privilege escalation"
+
+  parameters = <<PARAMS
+    {
+      "excludedNamespaces": {
+        "value": [
+          "kube-system",
```
sebader, line 33: so if you exclude any, custom namespaces, you also need to add the defaults I take it?!

PR #348, opened 2022-05-06
Various text improvements

msimecek:
When writing the tech update about user node pools I noticed a few things in the text which could be improved :)


heoelri (approved):
LGTM

PR #349, opened 2022-05-09
Microsoft Azure Load Test (MALT) Integration

heoelri:
This PR (re-)introduces support for Microsoft Azure Load Test (MALT). 

MALT will be added as a secondary load testing framework next to the existing Locust implementation. The main use-case for MALT is the embedded load testing in E2E. It's embedded into the E2E pipeline and can be selected at pipeline execution:

image

Embedded load testing can be enabled / disabled by checking / unchecking "Run load testing" the load testing framework is selected via "Select load testing framework". 

image

The selection made here also applies to "Run Chaos testing" which also contains a load test run while executing chaos experiments. 

The load test is using a JMX file stored in `/src/testing/loadtest-azure/scripts/catalogue-test.jmx` as well as some load test-related parameters loaded from:

https://github.com/Azure/Mission-Critical-Online/blob/fdba4f9edf67e9468c4f25e5797032dc2220122c/.ado/pipelines/config/configuration.yaml#L57-L67

Environment-specific parameters like the Front Door FQDN are set via parameters. The MALT service resource is deployed via Terraform, the pipeline implementation uses PowerShell-based wrapper scripts to interact with the MALT data plane API.

* Terraform: `/src/testing/loadtest-azure/infra`
* Wrapper Scripts: `/src/testing/loadtest-azure/scripts`

image

image

image

image

image


heoelri:
Merged into `feature/chaos-may-22`. Closed.

PR #351, opened 2022-05-09
Separate ingress definition from workload

sebader:
Adds the ingress definitions (which services to expose at which path) to a separate helm chart. This makes it easier to a) adopt the RI for other workloads and b) to add more services that should be expose (instead of just catalogservice and healthservice).

This also requires that each service is aware of the specific path-prefix, otherwise this requires a lot of rewrite-magic at each reverse proxy stage which in some cases doesn't even properly work at all. Thus the changes on the application code side to add the base path there.

It also enables to expose each swagger UI (when enabled). Currently running here:

https://afe2edd5b-global-fd.azurefd.net/
https://afe2edd5b-global-fd.azurefd.net/catalogservice/swagger
https://afe2edd5b-global-fd.azurefd.net/healthservice/swagger

@heoelri @msimecek thoughts?


heoelri:
I thought about the pros and cons and I'd advocate to **not** separate the ingress configuration into its own helm chart. I'd instead suggest to put the ingress rule back into each service's helm chart like what we had in the beginning:
*  [gameservice/templates/ingress.yaml](https://github.com/Azure/AlwaysOn-Engineering/blob/main/src/app/charts/gameservice/templates/ingress.yaml)
* [healthservice/templates/ingress.yaml](https://github.com/Azure/AlwaysOn-Engineering/blob/d8c4585f716838c2b6f9a4968280d323150edbd6/src/app/charts/healthservice/templates/ingress.yaml)

I think that this is the most reasonable and understandable approach. You're looking for the healthservice's ingress rule? It's in the healthservice helm chart. Packaging together k8s manifests and resources that share the same lifecycle. They're installed, updated and removed together.

```bash
- chart
| `- values.yaml
| `- Chart.yaml
| `- templates
|    `- deployment.yaml
|    `- ingress.yaml
|    `- hpa.yaml
|    `- ...
```
Why have we removed the ingress rule from the healthservice at some point (see [ingress.yaml](https://github.com/Azure/AlwaysOn-Engineering/blob/main/src/app/charts/healthservice/templates/ingress.yaml)) and moved it into the gameservice helmchart? The issue we were facing is that having an ingress rule in healthservice and gameservice caused cert-manager to request two certs for the same domain name, this caused a lot of conflicts and we often reached [Let's Encrypt's rate limits](https://letsencrypt.org/docs/rate-limits/). To work around the issue we've moved the healthservice ingress rule to the gameservice (later catalogservice):

https://github.com/Azure/Mission-Critical-Online/blob/fef64c0242db6799e1197a0beb981abb3876777b/src/app/charts/healthservice/templates/ingress.yaml#L1-L3

https://github.com/Azure/Mission-Critical-Online/blob/fef64c0242db6799e1197a0beb981abb3876777b/src/app/charts/catalogservice/templates/ingress.yaml#L56-L62

This approach has of course its own downsides - the healthservice now requires the catalogservice helm chart to be installed to expose its `/health` endpoint. This is indeed suboptimal.

That's why I'd suggest to go back to the initial approach - individual ingress rule definition per service / per helm chart. All (or multiple) charts can point to the same k8s secret that contains the TLS cert - and we can use an annotation to let cert-manager only request new certs for one of them.

```bash
kubectl get ingress -n workload
NAME                     CLASS   HOSTS                                                ADDRESS         PORTS     AGE
catalogservice-ingress   nginx   afe2e5465-cluster.swedencentral.cloudapp.azure.com   20.240.10.221   80, 443   10m
healthservice-ingress    nginx   afe2e5465-cluster.swedencentral.cloudapp.azure.com   20.240.10.221   80, 443   11m
```

```bash
kubectl describe ingress -n workload |grep -A1 TLS
TLS:
  workload-ingress-secret terminates afe2e5465-cluster.swedencentral.cloudapp.azure.com
--
TLS:
  workload-ingress-secret terminates afe2e5465-cluster.swedencentral.cloudapp.azure.com
```

Only the ingress rule deployed via the chart that has the `cert-manager.io/cluster-issuer: "cert-manager-alwayson"` annotation set will trigger cert-manager to request a new cert. 

https://github.com/Azure/Mission-Critical-Online/blob/4e004f768965e5fc8cc1608e97afaae4070c655a/src/app/charts/catalogservice/values.yaml#L17-L19

This is the same approach other projects with larger numbers of helm charts use. Like for example the [TailwindTraders](https://github.com/microsoft/TailwindTraders-Backend/tree/main/Deploy/helm) project on GitHub. 

See https://github.com/Azure/Mission-Critical-Online/pull/356 how the implementation could look like. And https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=1374&view=results for a successful test.

heoelri:
Couple of additional thoughts why I suggest to keep the Ingress rule in each service's helm chart instead of separating them into an "ingress" helm chart:

* When a service changes its path like for example from `/api/catalogservice` to `/api/gameservice`.

or
* In a newer release an additional path i.e. `/api/catalogservice/v2` is introduced that points to a different K8s service.

These changes are done in the per-service helm chart only. A newer version of the chart contains the change and can be rolled back the same way. The chart can be tested and reviewed and does not have dependencies to other (in this case a central ingress helm chart) charts.

When services are split to different clusters like for example not all services are deployed to all stamps/regions or within a given region there are frontend/backend clusters there's no need to then have multiple versions of this "central ingress chart".

Other benefits are that the risk that the "central ingress chart" contains abandoned paths is lower and changes to ingress affect only each individual chart - instead of changes made to a "central chart" that could potentially have an impact to all other services.

Changes to the "central ingress chart" would require testing ALL other charts relying/depending on this chart.

msimecek:
After reading the discussion and looking at the two variants of code, I prefer Heyko's approach to this one. Having the ingress config sitting together with each service means that changes to all ingresses need to be done in multiple places, however it facilitates the idea of common lifecycle and logical grouping.

heoelri:
> > Having the ingress config sitting together with each service means that changes to all ingresses need to be done in multiple places
> 
> what do you mean by this?
> 
> My concern with the decentralized approach is that you would need to bring infra-specifics into the applications - or find a different way to keep them out of their ingress definitions. Like the hostname part of the ingress file or additional header validations with front door which we are currently doing. I just doubt this is feasible to tell each app team "please incorporate these parts into your ingress definition"

I share your concerns, we've currently some very specific stuff in our ingress rules.

https://github.com/Azure/Mission-Critical-Online/blob/0a0ed298248a2cbb18f3fbdd34b172df7fa7d6ed/src/app/charts/catalogservice/templates/ingress.yaml#L10-L30

Some of these settings could potentially either go into the global ingress-nginx configuration, though things like `force-ssl-redirect` and others will probably not work with the currently used http01 certificate resolver (which requires HTTP), or could potentially being dropped like the Front Door Header in the AFDv2 with PL scenario.

To reduce the complexity we should think about ways to simplify these ingress rules / helm charts by moving/removing some of these annotations while still keeping them in each individual per-service helm chart.

msimecek:
> My concern with the decentralized approach is that you would need to bring infra-specifics into the applications - or find a different way to keep them out of their ingress definitions. Like the hostname part of the ingress file or additional header validations with front door which we are currently doing. I just doubt this is feasible to tell each app team "please incorporate these parts into your ingress definition"

First of all, I think it's okay for the app team to own their part of the system end to end. Second, no one is preventing the infra team from doing these changes if asking the owning app team is not feasible (due to the sheer number  of teams or whatever other reasons).

heoelri:
@sebader how do we want to proceed with this? Do we want to have another conversation about this? 

Or do we proceed with PR #356? 

cc: @msimecek

PR #355, opened 2022-05-10
Add timespan cutoff for health model queries

nielsams:
Ideally we would run the health model query over all the data we have, but after a few days these queries become to heavy to run in a time that is useful for visualization. 

On the long run, we can either
- Dive further into Log Analytics optimizations to see if we can run the queries more efficiently
- Move to a model where the health queries are pre-calculated and only visualized instead of calculated on the fly

For now, we're adding a cutoff for ago(1d), meaning the health model only returns scores for the last 24 hours.


.ado/pipelines/config/variables-values-e2e.yaml
```diff
@@ -9,7 +9,7 @@ variables:
 # The first value in 'stampLocations' is the primary region used for global services.
 # IMPORTANT! Changing the primary region (first value) is a BREAKING change and will destroy CosmosDB and Front Door.
 - name: 'stampLocations'
-  value: '["swedencentral"]'    # Check which regions are valid. There is a list in /src/infra/README.md
+  value: '["brazilsouth"]'    # Check which regions are valid. There is a list in /src/infra/README.md
```
sebader, line 12: please revert before merging

src/infra/monitoring/queries/stamp/BackgroundProcessorHealthStatus.kql
```diff
@@ -7,7 +8,7 @@ let Thresholds=datatable(MetricName: string, YellowThreshold: double, RedThresho
     "avgProcessingTime", 100, 200
     ];
 //
-let timespanStart= todatetime(format_datetime( todatetime(toscalar(AppRequests| summarize min(TimeGenerated))), 'yyyy-MM-dd HH:ss'));
+let timespanStart = _timespanStart;//todatetime(format_datetime( todatetime(toscalar(AppRequests| summarize min(TimeGenerated))), 'yyyy-MM-dd HH:ss'));
```
sebader, line 11: do you foresee ever switching back to the old code? if not, I would just remove it

sebader (changes requested):

src/infra/monitoring/queries/stamp/BackgroundProcessorHealthStatus.kql
nielsams: Yes, I hope we can move back to that soon. Would be great if we can control the timespan elsewhere.

sebader (approved):

PR #307, opened 2022-04-25
Simplify helm chart installation

heoelri:
Helm 3.x does now finally support the creation of not existing namespaces, again this was already the case in Helm 2.x.

This PR simplifies the installation via `helm upgrade` by using `--create-namespace` to create the target namespace (if not exist) w/o the need for an additional `kubectl apply` statement.

**DRAFT** - Waiting for proper testing till the issue with the ingress path was fixed. https://helm.sh/docs/helm/helm_install/

PR #356, opened 2022-05-10
Individual Ingress rules per service/helm chart

heoelri:
This PR splits the single ingress rule that exposes the `catalogservice` and the `healthservice` that currently exist only in the `catalogservice` helm chart into individual ingress rules per service, packaged together with the service in individual helm charts. See discussion in PR #351 for more details.

### DIFF
--- a/.ado/pipelines/templates/jobs-workload-deploy.yaml
+++ b/.ado/pipelines/templates/jobs-workload-deploy.yaml
@@ -44,8 +44,11 @@ jobs: # using jobs so they each run in parallel
             echo  "*** Deploy healthservice to $aksClusterName (in $aksClusterResourceGroup) via helm chart"
             helm upgrade --install healthservice-workload src/app/charts/healthservice `
                           --namespace $(workloadNamespace) --create-namespace `
-                          --set containerimage="$fullContainerImageName" `
+                          --set azure.frontdoorid="$frontDoorId" `
                           --set azure.region="$region" `
+                          --set azure.baseurl="$frontDoorFqdn" `
+                          --set containerimage="$fullContainerImageName" `
+                          --set workload.domainname="$fqdn" `
                           ${{ parameters.additionalHelmArguments }}
 
             # Wait for healthservice-deploy k8s deployment to come up
--- a/.ado/pipelines/templates/jobs-workload-deploy.yaml
+++ b/.ado/pipelines/templates/jobs-workload-deploy.yaml
@@ -25,6 +25,14 @@ jobs: # using jobs so they each run in parallel
           # load json data from downloaded terraform artifact
           $releaseUnitInfraDeployOutput = Get-ChildItem $(Pipeline.Workspace)/terraformOutputReleaseUnitInfra/*.json | Get-Content | ConvertFrom-JSON
 
+          $globalInfraDeployOutput = Get-ChildItem $(Pipeline.Workspace)/terraformOutputGlobalInfra/*.json | Get-Content | ConvertFrom-JSON
+
+          $frontDoorId = $globalInfraDeployOutput.frontdoor_id_header.value
+          echo "*** Retrieved Azure Front Door Header ID $frontDoorId"
+
+          $frontDoorFqdn = $globalInfraDeployOutput.frontdoor_fqdn.value
+          echo "*** Retrieved Azure Front Door FQDN $frontDoorFqdn"
+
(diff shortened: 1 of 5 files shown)


heoelri:
One example from a recent sample implementation:

image

Each API has its own helm chart with its own ingress(.yaml) definition.

image

They all use the same ingress controller / fqdn and all of them share the same certificate:

image

image

heoelri:
The current schema here is `/catalogservice/api/*` and `/healthservice/api/*` wouldn't it be easier if we'd switch to `/api/healthservice/*` and `/api/catalogservice/*`.

heoelri:
> The current schema here is `/catalogservice/api/*` and `/healthservice/api/*` wouldn't it be easier if we'd switch to `/api/healthservice/*` and `/api/catalogservice/*`.

@sebader do we want to consolidate all APIs under `/api` as part of this PR? Currently all APIs are hosted on AKS and this would reduce the need to modify the FD config when APIs are changed, removed or added. Like in the diagram shared above:

image

msimecek:
Removing `/api` from the default route is an easy thing and it will make things more logical (which I realized last week too, trying to figure out what the correct route is without looking :D). So I support that.

sebader:
> Removing `/api` from the default route is easy and it will make things more logical (which I realized last week too, trying to figure out what the correct route is without looking :D). So I support that.

hm, I just started to do that in the code when I realized: It actually is not that easy. Because then we are back in reverse procxy rewrite hell :(

.ado/pipelines/templates/jobs-workload-deploy.yaml
```diff
@@ -27,9 +27,6 @@ jobs: # using jobs so they each run in parallel
           $frontDoorId = $globalInfraDeployOutput.frontdoor_id_header.value
           echo "*** Retrieved Azure Front Door Header ID $frontDoorId"
 
-          $frontDoorFqdn = $globalInfraDeployOutput.frontdoor_fqdn.value
```
heoelri: This shouldn't exist anymore. It should have been dropped in PR #307.

src/app/charts/catalogservice/values.yaml
```diff
@@ -24,7 +25,3 @@ networkPolicy:
 azure:
   frontdoorid: "" # Azure Front Door Header ID (blank = disabled)
```
heoelri, line 26:
```suggestion
  frontdoorid: "" # Azure Front Door Header ID
```

src/app/charts/healthservice/values.yaml
```diff
@@ -17,4 +22,5 @@ networkPolicy:
   enabled: true # Enable Network Policy
 
 azure:
+  frontdoorid: "" # Azure Front Door Header ID (blank = disabled)
```
heoelri, line 25:
```suggestion
  frontdoorid: "" # Azure Front Door Header ID
```

src/app/charts/catalogservice/values.yaml
```diff
@@ -15,7 +16,7 @@ scale: # Horizontal Pod Autoscaler (HPA) configuration
 
 ingress:
   annotations:
-    cert-manager.io/cluster-issuer: "cert-manager-alwayson"
+    cert-manager.io/cluster-issuer: "cert-manager-alwayson" # this will trigger cert-manager to request a certificate
```
sebader, line 19: I'm thinking we should move this to the healthservice instead:
- we'll expect many customers might reuse our healthservice anyway. Then they dont need to move this part around for their own workload and decide on which one to annotate
- healthservice currently gets installed first anyway

heoelri: Yes, I agree. That makes sense.

.ado/pipelines/templates/jobs-workload-deploy.yaml
sebader: good spot. maybe was some merge conflict. remove it

src/app/charts/catalogservice/values.yaml
sebader: ok done

heoelri:
Seems to work.. https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=1547&view=results

image

heoelri:
> lets just once more deploy it on a fresh branch so that we get a brand new cert

before we do that let me parameterize a few more things

PR #381, opened 2022-05-11
CosmosDB updates

esbran:
Aditions to CosmosDB design


sebader:
ok, looking at this PR I would say this actually all should go int WAF/MS Docs :) It does not describe any actual decisions taken for the Reference Implementation but general considerations when using Cosmos.

really good content, though!

esbran:
Thank you Sebastian, I will move it to MS-docs.

PR #382, opened 2022-05-11
Remove retry code

sebader:
Removes unused retry code in two places:

- terraform apply: Cannot be retried on fail since it will error out with "plan stale"
image

- No wait required on rollout of backgroundprocessor and healthservice since we are already using helm --wait there

image


heoelri:
I think I'm ok with that - I haven't seen this retry logic to kick in for a while now. 

The `--wait` parameter for helm should be sufficient to replace our wait loop:

> --wait: Waits until all Pods are in a ready state, PVCs are bound, Deployments have minimum (Desired minus maxUnavailable) Pods in ready state and Services have an IP address (and Ingress if a LoadBalancer) before marking the release as successful. It will wait for as long as the --timeout value. If timeout is reached, the release will be marked as FAILED. Note: In scenarios where Deployment has replicas set to 1 and maxUnavailable is not set to 0 as part of rolling update strategy, --wait will return as ready as it has satisfied the minimum Pod in ready condition.
> https://helm.sh/docs/intro/using_helm/#helpful-options-for-installupgraderollback

We should keep parts of the retry logic for the catalogservice though cause it's checking for the certificate secret that's created by cert-manager and not by the helm chart:

https://github.com/Azure/Mission-Critical-Online/blob/574e7124cc0b048203f23ae349eb7057f620d725/.ado/pipelines/templates/jobs-workload-deploy.yaml#L132-L153

sebader:
@heoelri thats why I didnt remove the retry logic for the catalogservice. or what are you missing?

heoelri:
> @heoelri thats why I didnt remove the retry logic for the catalogservice. or what are you missing?

Yes, I saw that. Nothing is missing. Approved.

heoelri (approved):

heoelri:
Will you port this to -connected as well? 

https://github.com/Azure/Mission-Critical-Connected/blob/802ff2b617d5f266bd8b4bf5dabe17783e6b211c/.ado/pipelines/templates/jobs-workload-deploy.yaml#L51-L65

PR #383, opened 2022-05-12
API paths per service

sebader:
This also requires that each service is aware of the specific path-prefix, otherwise this requires a lot of rewrite-magic at each reverse proxy stage which in some cases doesn't even properly work at all. Thus the changes on the application code side to add the base path there.

It also enables to expose each swagger UI (when enabled). Currently running here:

https://afe2edd5b-global-fd.azurefd.net/
https://afe2edd5b-global-fd.azurefd.net/catalogservice/swagger
https://afe2edd5b-global-fd.azurefd.net/healthservice/swagger

### DIFF
--- a/.ado/pipelines/README.md
+++ b/.ado/pipelines/README.md
@@ -91,10 +91,10 @@ All pipelines are using Azure DevOps service connections to connect to Microsoft
 
 As part of the pipelines, basic Smoke Tests against the APIs are executed:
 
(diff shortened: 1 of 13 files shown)


sebader:
validated in e2e

msimecek (approved):
LGTM. I assume you searched for these API calls across full codebase :)

sebader:
I assume so, too. And I assume that you as a reviewer did the same :D

PR #384, opened 2022-05-12
Various chaos and load test configuration updates

msimecek:
Bringing over the timings from the chaos demo to `main`.

Tested here: https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=1405&view=results

In the load test run (which is not visible anymore) the timings were a bit off (test was shorter than expected, only 1 failure etc.), but the configuration was applied from the pipeline, so it might be something with the service or pipeline synchronization.

### DIFF
--- a/.ado/pipelines/templates/stages-chaos.yaml
+++ b/.ado/pipelines/templates/stages-chaos.yaml
@@ -8,10 +8,8 @@ parameters:
   - name: customPrefix
     type: string
   - name: testDurationSeconds
-    type: number
     default: 0
-  - name: loadTestStartSeconds
-    type: number
+  - name: startDelay
     default: 300
   - name: podFailure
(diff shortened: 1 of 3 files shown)


.ado/pipelines/templates/stages-chaos.yaml
```diff
@@ -8,10 +8,8 @@ parameters:
   - name: customPrefix
     type: string
   - name: testDurationSeconds
-    type: number
     default: 0
-  - name: loadTestStartSeconds
-    type: number
```
msimecek, line 14: I had to remove types in order to use variable values when calling this template... Locust load test has the same thing.

sebader (changes requested):

sebader (approved):

PR #391, opened 2022-05-13
Drop inline parsing of global tf outputs

heoelri:
This PR replaces the in some parts of the pipeline used inline parsing of the global terraform deployment outputs with the automated parsing and publishing (as pipeline vars) via `steps-parse-terraform-output.yaml`.

```yaml
- template: steps-parse-terraform-output.yaml # parse global configuration settings from terraform deployment output
  parameters:
    workingDirectory: '$(Pipeline.Workspace)/terraformOutputGlobalInfra'  # Global infra deploy output directory
```

image

### DIFF
--- a/.ado/pipelines/templates/jobs-workload-deploy.yaml
+++ b/.ado/pipelines/templates/jobs-workload-deploy.yaml
@@ -58,6 +58,10 @@ jobs: # using jobs so they each run in parallel
 
     - template: steps-buildagent-prerequisites.yaml
 
+    - template: steps-parse-terraform-output.yaml
+      parameters:
+        workingDirectory: '$(Pipeline.Workspace)/terraformOutputGlobalInfra'  # Global infra deploy output directory
+
     - task: AzureCLI@2
       displayName: 'Install workload CatalogService on AKS clusters'
       retryCountOnTaskFailure: 2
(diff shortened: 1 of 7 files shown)


.ado/pipelines/templates/jobs-workload-deploy.yaml
```diff
@@ -182,7 +179,7 @@ jobs: # using jobs so they each run in parallel
             # Apply workload BackgroundProcessor helm chart
             echo  "*** Deploy workload BackgroundProcessor to $aksClusterName (in $aksClusterResourceGroup) via helm chart"
             helm upgrade --install workload-backgroundprocessor src/app/charts/backgroundprocessor `
-                         --namespace $(workloadNamespace) --create-namespace `
+                         --namespace $(workloadNamespace) --create-namespace --wait `
```
sebader, line 182: --wait comes already from this (although I would actually vote to drop the parameter since we are never using it and just always specify --wait where we want it

parameters:
  additionalHelmArguments:     '--wait'

sebader (changes requested):
I guess I'm ok with this change

.ado/pipelines/templates/jobs-workload-deploy.yaml
heoelri: cannot remember why we've introduced `additionalHelmArguments` in the first place. It's kind of useless here anyhow given how different the workloads deployed in `jobs-workload-deploy.yaml` are. It's not really an abstract and reusable template like most of the others.

sebader: yeah then just remove it

heoelri:
Tested it in E2e: https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=1415&view=results

.ado/pipelines/templates/jobs-workload-deploy.yaml
```diff
@@ -45,8 +42,7 @@ jobs: # using jobs so they each run in parallel
             helm upgrade --install healthservice-workload src/app/charts/healthservice `
                           --namespace $(workloadNamespace) --create-namespace `
                           --set containerimage="$fullContainerImageName" `
-                          --set azure.region="$region" `
-                          ${{ parameters.additionalHelmArguments }}
+                          --set azure.region="$region"
```
sebader, line 45: this should get --wait now

.ado/pipelines/templates/jobs-workload-deploy.yaml
```diff
@@ -107,12 +102,11 @@ jobs: # using jobs so they each run in parallel
...
                          --namespace "$(workloadNamespace)" --create-namespace `
-                         --set azure.frontdoorid="$frontDoorId" `
+                         --set azure.frontdoorid="$(frontdoor_id_header)" `
                          --set azure.region="$region" `
-                         --set azure.baseurl="$frontDoorFqdn" `
+                         --set azure.baseurl="$(frontdoor_fqdn)" `
                          --set containerimage="$fullContainerImageName" `
-                         --set workload.domainname="$fqdn" `
-                         ${{ parameters.additionalHelmArguments }}
+                         --set workload.domainname="$fqdn"
```
sebader, line 109: also --wait

sebader (changes requested):

sebader (approved):
ok for me.
@msimecek thoughts?

heoelri:
> image

The only thing i thought about was if it makes sense to add a `global_` or `tf_` prefix to these variables to make clear that these are pipeline variables and where they're coming from.

sebader:
good point. Maybe we could add tf_ but not sure if we are not fully relying on the names being the same as the terraform outputs somewhere?

PR #396, opened 2022-05-17
Update HPA to v2

sebader:
Update horizontal pod autoscalers to v2

@heoelri

### DIFF
--- a/src/app/charts/backgroundprocessor/templates/hpa.yaml
+++ b/src/app/charts/backgroundprocessor/templates/hpa.yaml
@@ -1,4 +1,4 @@
-apiVersion: autoscaling/v1
+apiVersion: autoscaling/v2
 kind: HorizontalPodAutoscaler
 metadata:
   name: {{ .Chart.Name }}-autoscaler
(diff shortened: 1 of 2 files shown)


heoelri:
Bumping `kubeVersion` from `>=1.21.0` to `>=1.23.0` given that `autoscaling/v2` is brandnew in K8s `1.23` and not available in older versions.

https://github.com/Azure/Mission-Critical-Online/blob/20975f31288077a8b9780a1e1a55942b79d316e8/src/app/charts/catalogservice/Chart.yaml#L8

https://github.com/Azure/Mission-Critical-Online/blob/20975f31288077a8b9780a1e1a55942b79d316e8/src/app/charts/backgroundprocessor/Chart.yaml#L8

https://github.com/Azure/Mission-Critical-Online/blob/20975f31288077a8b9780a1e1a55942b79d316e8/src/app/charts/healthservice/Chart.yaml#L8

sebader:
good catch!

heoelri (approved):
I haven't had a chance to work with `autoscaling/v2` yet - I assume it works as expected, given that it was in beta before, so ok from my side. 

But maybe we can test it once before merging it, by running a load test against a deployment to see if the workload pods are still scaling in and out as expected.

heoelri:
> the env got already destroyed. but feel free to run again :) (I also just trusted it that it does work and scale out...)

I agree.. 

> The v2 of the Horizontal Pod Autoscaler (HPA) API has been around since November, 2016. Five years ago! Although it is pretty much stable, there are some minor issues before it can be officially considered as so.
> https://sysdig.com/blog/kubernetes-1-23-whats-new/

sebader:
works
image

PR #398, opened 2022-05-25
Move to AKS OMS container log schema v2

sebader:
This brings a lot more information directly to the relevant log entry, like which namespace and pod:
image

https://azure.microsoft.com/en-us/updates/public-preview-containerlogv2-schema-in-azure-monitor-container-insights/


heoelri (approved):
LGTM

PR #403, opened 2022-05-31
Cleanup and optimize Helm charts

heoelri:
This PR superseeds PR #356 and #351. See #356 for more details. 

https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=1552&view=results


src/app/charts/backgroundprocessor/templates/deployment.yaml
```diff
@@ -35,13 +34,9 @@ spec:
...
           initialDelaySeconds: 5
           periodSeconds: 10
-        resources:
-          limits:
-            cpu: "0.5"
-            memory: "1Gi"
-          requests:
-            cpu: "0.5"
-            memory: "512Mi"
+        {{- if .Values.resources }}
```
sebader, line 37: should this even have an if? When this is not specified, deployment should fail

heoelri: The currently available policies do not really solve that. You can specify limits, but you cannot make `resources` mandatory. For now I'd like to keep it as it is. Our helm charts are considered to follow "best practices" and I'd not rely on any existing or not existing expressions inside the helm chart to fail a deployment.

Especially as no one would prevent a user from removing resources from the `values.yaml` as well as from the `templates/deployment.yaml`.

heoelri:
Needs to be ported to connected as well.

.ado/pipelines/templates/jobs-workload-deploy.yaml
```diff
@@ -74,18 +107,15 @@ jobs: # using jobs so they each run in parallel
           # load json data from downloaded terraform artifact
           $releaseUnitInfraDeployOutput = Get-ChildItem $(Pipeline.Workspace)/terraformOutputReleaseUnitInfra/*.json | Get-Content | ConvertFrom-JSON
 
-          echo "*** Retrieved Azure Front Door Header ID $(frontdoor_id_header) from Terraform output"
```
heoelri, line 77: @sebader was it intended to drop this? I think this is valuable debug output.

sebader: yeah I realized we never really needed this one (or never had issues). you can always just look it up in the artifact. and it is kind of a secret so not logging it is probably better

PR #423, opened 2022-06-02
Fail on ingress certificate never getting ready

sebader:
Attempt to properly fail the pipeline when the cert is never ready. Also extended the timeout. Worked in e2e but we will only see over time if it actually helps

https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=1579&view=results


heoelri (approved):
ok for me

sebader:
> > ok for me
> 
> not sure though if this will really fix the issue and if its just a question of time..

yeah not 100% sure either but over the weekend we had another failed pipeline. Rerunning the task now worked since the cert was provisioned now

PR #427, opened 2022-06-03
Improvements on health queries

sebader:
This should fix a couple of metrics like PodRestart to show more meaningful data.
Most importantly it should also bring a huge performance improvement as now every query includes the timefilter for two days instead of partially scanning all 30 days of data.

More changes around the queries will follow over the next weeks.

@nielsams


heoelri (approved):
fingers crossed

PR #430, opened 2022-06-15
Small infrastructure improvements

sebader:
- Increase AFD latency to 1s to cater for the very distributed deployment of INT (brazil and sweden right now)
- Show how to enable DDoS protection which is recommended to use in production

### DIFF
--- a/src/infra/workload/releaseunit/modules/stamp/network.tf
+++ b/src/infra/workload/releaseunit/modules/stamp/network.tf
@@ -24,6 +24,13 @@ resource "azurerm_virtual_network" "stamp" {
   location            = azurerm_resource_group.stamp.location
   address_space       = [module.subnet_addrs.base_cidr_block]
 
+  # For production workloads we recommend to enable DDoS protection Standard plan
+  # You can link this to an existing DDoS protection plan or create a new one
+  # ddos_protection_plan {
+  #   id     = var.ddos_protection_plan_id
+  #   enable = true
+  # }
+
   tags = var.default_tags
 }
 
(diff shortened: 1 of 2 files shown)


src/infra/workload/releaseunit/modules/stamp/network.tf
```diff
@@ -24,6 +24,13 @@ resource "azurerm_virtual_network" "stamp" {
   location            = azurerm_resource_group.stamp.location
   address_space       = [module.subnet_addrs.base_cidr_block]
 
+  # For production workloads we recommend to enable DDoS protection Standard plan when exposing your AKS ingress on public IPs
+  # You can link this to an existing DDoS protection plan or create a new one
+  # ddos_protection_plan {
+  #   id     = var.ddos_protection_plan_id
```
msimecek, line 30: Where is this variable declared and how does one get the value of the procection plan?

msimecek (changes requested):

src/infra/workload/releaseunit/modules/stamp/network.tf
sebader: damn, you dont miss a thing :D I was wondering if I should actually include all that, too? would be all commented out, too. do you want me to?

msimecek: Maybe just explain what needs to be done for this to work? "add this variable to variables.tf, get the plan ID from..."

msimecek: Better :) Now I've learned that it's supposed to be resource ID.

msimecek (approved):

PR #466, opened 2022-07-07
Adding AKS user node pool

sebader:
In line with WAF recommendations, we should split into dedicated system and user node pools for AKS

- Adds a user node pool ("workload") and puts taints for nginx and the workload deployments to run there
- Uses smaller VM SKU for the system node pool. Possible by making the OS disk size smaller

@heoelri 

image

image
image

### DIFF
--- a/src/config/ingress-nginx/values.helm.yaml
+++ b/src/config/ingress-nginx/values.helm.yaml
@@ -22,7 +22,7 @@ controller:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/port: "10254"
-  tolerations:
+  tolerations: # Schedule the ingress controller on our user node pool
   - key: "workload"
     operator: "Equal"
     value: "true"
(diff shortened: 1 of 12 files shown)


src/config/ingress-nginx/values.helm.yaml
```diff
@@ -22,6 +22,11 @@ controller:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/port: "10254"
+  tolerations: # Schedule the ingress controller on our user node pool
```
heoelri, line 25: I wasn't sure about that, but I think I agree that it makes sense to host the ingress controller which is defacto used only to expose the workload, in the workload node pool as well.

sebader: yeah I think it makes a lot of sense, because this way you at least have a decent number of requests that will not require an additional network hop from ingress to the workload pods. Plus, the system load that the ingress controllers generate is directly linked to the overall request rate. Thus you probably want you ingress to not impact any of your system pods - just like with your actual workload pods

heoelri (approved):
LGTM

PR #468, opened 2022-07-19
Update MALT

heoelri:
This PR contains some significant updates to our Microsoft Azure Load Test (MALT) implementation.

Highlights:

* Updated API version from `2021-07-01-preview` to `2022-06-01-preview`
* Implementing the native "Test criteria" functionality

# Test criteria

Test criteria is a native functionality in MALT that allows to pre-define test criteria that will fail tests if any of the criteria evaluates to true.

image


heoelri:
Here's a first draft how the "test criteria" implementation could potentially look like. `loadtest-create.ps1` gets an additional parameter 
```powershell
[string] $passFailCriteria
```

The parameter is optional, test criteria are only set when this parameter is set:

```powershell
if ($passFailCriteria) {
  if (!(Test-Path $passFailCriteria)) {
    throw "ERROR: $passFailCriteria not found or invalid."
  }

  $content = Get-Content $passFailCriteria | ConvertFrom-JSON

  $jsonBase = $body | convertfrom-json  
  $jsonBase | Add-Member -MemberType NoteProperty -Name "passFailCriteria" -Value $content

  $body = $jsonBase | ConvertTo-JSON -Depth 4

  Write-Host $body
}
```

That allows us to handover a path to a JSON file containing our criteria definition. This definition could look like this:

```JSON
[
    {
        "passFailMetrics": {
            "ece127bc-6271-4a70-93f6-758c56a3f12c": {
                "clientmetric": "requests_per_sec",
                "aggregate": "avg",
                "condition": "<",
                "value": 1000.0,
                "result": null,
                "actualValue": 0.0,
                "requestName": null,
                "action": "continue"
            } 
        }
    }
]
```

This would allow us to handover different "baseline" JSON files for example for a regular test and for chaos testing (either one or different depending on the kind of tests selected - today we do not have an option to select different sets of experiments though).

Current issues: 
* The format for `passFailCritera` needs to be correct - otherwise the test creation will fail
* Each criteria needs a GUID

heoelri:
I think it would be good to show somewhere, for example in the test run displayname, that/when chaos was running in parallel.

.ado/pipelines/config/loadtest-baseline-chaos.json
```diff
@@ -0,0 +1,36 @@
+[
+    {
+        "passFailMetrics": {
+            "ece127bc-6271-4a70-93f6-758c56a3f11c": {
```
sebader, line 4: where do those guids(?) come from? self-generated?

.ado/pipelines/config/loadtest-baseline-normal.json
```diff
@@ -0,0 +1,36 @@
+[
+    {
+        "passFailMetrics": {
+            "ece127bc-6271-4a70-93f6-758c56a3f11c": {
```
sebader, line 4: and I guess they dont have to be unique?!

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -18,6 +18,11 @@ parameters:
    type: string
    default: ''
 
+ - name: runChaosTesting
+   displayName: "Run chaos testing"
```
sebader, line 22:
```suggestion
   displayName: "Run chaos testing" # Influences which baseline to compare to
```

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -38,7 +43,7 @@ stages:
         terraformStateFilename:            'terraform-azurelt-${{ parameters.customPrefix }}.state'
         terraformWorkingDirectory:         '${{ parameters.terraformWorkingDirectory }}'
 
-    # Deploy the locust infrastructure. If running in headless mode, the test will automatically start once the infra is provisioned.
+    # Deploy the MALT infrastructure. If running in headless mode, the test will automatically start once the infra is provisioned.
```
sebader, line 46:
```suggestion
    # Deploy the MALT infrastructure. If running in embedded mode, the test will automatically start once the infra is provisioned.
```

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -66,22 +71,35 @@ stages:
       - script: |
           echo "*** Setting loadTestTargetFQDN to $(frontdoor_fqdn)"
           echo "##vso[task.setvariable variable=azureLoadTestTargetFQDN]$(frontdoor_fqdn)"
-        displayName: 'Set Loadtest targetUrl to Azure Front Door'
 
-    # Set loadTestTargetFQDN to the FQDN specified in the pipeline parameters
+          echo "*** Setting loadTestBaseline to $(azureLoadTestBaseline)"
+          echo "##vso[task.setvariable variable=loadTestBaseline]$(azureLoadTestBaseline)"
+
+        displayName: 'Set Azure Load Test variables'
```
sebader, line 78: I'm not a big fan of this condensed task syntax - especially when display name now comes at the bottom. can you maybe rewrite to the full task?

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -66,22 +71,35 @@ stages:
...
+
+    # override baseline when chaos testing is enabled
+    - ${{ if eq(parameters.runChaosTesting, 'true') }}:
+      - script: |
+          echo "*** Setting loadTestBaseline to $(chaosLoadTestBaseline)"
+          echo "##vso[task.setvariable variable=loadTestBaseline]$(chaosLoadTestBaseline)"
+        displayName: 'Override loadTestBaseline for Chaos Experiments'
```
sebader, line 85: same here

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -66,22 +71,35 @@ stages:
...
+
+    # setting parameters for embedded load tests
```
sebader, line 87: ??
if ne(parameters.embeddedLoadTest, 'true')  ==> # setting parameters for embedded load tests

sebader: should it be "parameters for standalone load tests"?

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -138,7 +157,7 @@ stages:
 
       # Create config csv for Azure Load Test
       - task: PowerShell@2
-        displayName: 'Create config *.csv file'
+        displayName: 'Create config JMeter file'
```
sebader, line 160:
```suggestion
        displayName: 'Create JMeter config file'
```

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -152,7 +171,7 @@ stages:
 
       # upload a loadtest config file from previous task
       - task: AzureCLI@2
-        displayName: 'Upload auto-generated load test config file for Azure Load Test'
+        displayName: 'Upload auto-generated config file'
```
sebader, line 174: why auto-generated?

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -185,6 +204,7 @@ stages:
       # get load test status and wait till it's done
       - task: AzureCLI@2
         displayName: 'Wait for load test to be finished'
+        continueOnError: true # shows a warning but continues the pipeline
```
sebader, line 207: is that for the failed baseline comparison? maybe add that. But I guess it also means that the pipeline would continue if something else breaks?!

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -208,7 +229,13 @@ stages:
...
                 Invoke-WebRequest $logsUrl -OutFile "results/logs.zip"
+
+                if ($result.testResult -eq "FAILED") {
+                  throw "*** ERROR: Test $($result.testRunId) failed."
+                }
+
               }
+
               # todo - timeout?
             } while ($result.status -ne "DONE")
```
sebader, line 240: are we sure there is no other end state except for DONE? I had that somewhere else already that it would stop with ERROR, or something else, then the pipeline would get stuck forever

.ado/pipelines/templates/stages-loadtest-locust.yaml
```diff
@@ -9,20 +9,11 @@ parameters:
    default: ''
```
sebader, line 9: as discussed: how about putting this directly into the pipeline file? that is the only place it is now referenced, right?

src/testing/README.md
```diff
@@ -24,18 +24,18 @@ Azure Mission-Critical integrates [Azure Chaos Studio](https://aka.ms/chaosstudi
...
+- [Azure Load Test Service](#azure-load-test-service) for load testing using the Microsoft Azure Load Test service
+- [Locust](#locust) for load testing using the open source load testing framework locust
 - [Playwright](#playwright) for UI testing
 - [Azure Chaos Studio](#azure-chaos-studio) for failure injection testing
 
-### JMeter with Azure Load Test
+### Azure Load Test Service
 
-[Azure Load Test](https://docs.microsoft.com/azure/load-testing/overview-what-is-azure-load-testing) is a managed service to execute [Apache JMeter](https://jmeter.apache.org/) load test definitions.
+Microsoft [Azure Load Test](https://docs.microsoft.com/azure/load-testing/overview-what-is-azure-load-testing) Service (MALT) is a managed service to execute [Apache JMeter](https://jmeter.apache.org/) load test definitions. Azure Mission-Critical comes with a load test definition for its sample application that can be leveraged either standalone (infrastructure and test service is deployed via a separate pipeline) or embedded into the e2e pipeline including a baseline definition.
```
sebader, line 34: I think MALTs plan is still to add more frameworks later. So maybe rewrite to say "currently its supporting JMeter, more to follow" etc?

src/testing/README.md
```diff
@@ -24,18 +24,18 @@ Azure Mission-Critical integrates [Azure Chaos Studio](https://aka.ms/chaosstudi
...
 
 ### Locust
 
-Locust is an open source Load Testing framework written in Python. See [locust](./loadtest-locust/README.md) for more details about the implementation and configuration.
+Locust is an open source Load Testing framework written in Python. See [locust](./loadtest-locust/README.md) for more details about the implementation and configuration. Locust was used before MALT was available and is still available as a [standalone deployment using Azure Container Instances](https://medium.com/microsoftazure/globally-distributed-load-tests-in-azure-with-locust-aeb3a365cd60).
```
sebader, line 38:
```suggestion
Locust is an open source Load Testing framework written in Python. See [locust](./loadtest-locust/README.md) for more details about the implementation and configuration. Locust was used in the Mission-Critical reference implementation before MALT was available and is still available as a [standalone deployment using Azure Container Instances](https://medium.com/microsoftazure/globally-distributed-load-tests-in-azure-with-locust-aeb3a365cd60).
```

src/testing/loadtest-azure/README.md
```diff
@@ -0,0 +1,33 @@
+# Load testing with Azure Load Test Service (MALT)
+
+Azure Mission-Critical uses the Microsoft Azure Load Test Service (MALT) for load testing. MALT is a Apache JMeter-based managed service and is implemented as a standalone as well as as a pipeline-embedded solution in Azure Mission-Critical.
```
sebader, line 3: same here regarding JMeter

src/testing/loadtest-azure/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Standalone creates the Load test infrastructure, creates a test and uploads the JMeter load test definition created and optimized for the Azure Mission-Critical sample application and its APIs.
```
sebader, line 5:
```suggestion
The Standalone pipeline creates the Load test resource in Azure, creates a test and uploads the JMeter load test definition created and optimized for the Azure Mission-Critical sample application and its APIs.
```

src/testing/loadtest-azure/scripts/loadtest-create.ps1
```diff
@@ -68,9 +70,27 @@ function GetTestBody {
...
+if ($passFailCriteria) {
+  Write-Verbose "*** passFailCriteria set to $passFailCriteria"
+
+  if (!(Test-Path $passFailCriteria)) {
+    throw "ERROR: $passFailCriteria not found or invalid."
+  }
+
+  $content = Get-Content $passFailCriteria | ConvertFrom-JSON
+
+  $jsonBase = $body | convertfrom-json
```
sebader, line 87:
```suggestion
  $jsonBase = $body | ConvertFrom-Json
```

sebader (changes requested):

src/testing/loadtest-azure/README.md
```diff
@@ -0,0 +1,33 @@
+# Load testing with Azure Load Test Service (MALT)
+
+Azure Mission-Critical uses the Microsoft Azure Load Test Service (MALT) for load testing. MALT is a Apache JMeter-based managed service and is implemented as a standalone as well as as a pipeline-embedded solution in Azure Mission-Critical.
+
+The Standalone pipeline creates the Load test resource in Azure, creates a test and uploads the JMeter load test definition created and optimized for the Azure Mission-Critical sample application and its APIs.
```
heoelri, line 5:
```suggestion
The **Azure.AlwaysOn Deploy Azure Load Test (standalone)** pipeline creates the Load test resource in Azure, creates a test and uploads the JMeter load test definition created and optimized for the Azure Mission-Critical sample application and its APIs.
```

.ado/pipelines/templates/stages-loadtest-azure.yaml
heoelri: Not 100% sure and the API is still preview and the potential values for `status` are not further documented (see [here](https://docs.microsoft.com/en-us/rest/api/loadtesting/dataplane/test-run/get-test-run?tabs=HTTP)).  But it was properly tested. The test result itself is not in `$result.status` its in `$result.testResult`.

.ado/pipelines/config/loadtest-baseline-normal.json
heoelri: Not clear. Needs further investigation. I'd prefer to auto-generate them instead of defining them in our baseline json at some point.

.ado/pipelines/templates/stages-loadtest-azure.yaml
heoelri: it was autogenerated in an earlier version but currently not used anymore - dropping the whole task now

.ado/pipelines/templates/stages-loadtest-azure.yaml
heoelri: I don't have a strong preference and do not insist on keeping it like that. Will change it to the more prescriptive task syntax.

src/testing/README.md
heoelri: We can do that. For now its only JMeter and all we do is build around that. Not sure if/when we'll see other frameworks and if that's a top priority for them.

.ado/pipelines/config/loadtest-baseline-normal.json
sebader: maybe just make sure they are at least unique in our two baseline files. no need to generate them on the fly in the pipeline, I guess

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -138,7 +157,7 @@ stages:
 
       # Create config csv for Azure Load Test
```
sebader, line 158: should we then just drop this task, too?

heoelri: was already dropped just not synced

.ado/pipelines/templates/stages-loadtest-azure.yaml
sebader: ok, I'm fine to go with this. Otherwise we would just need to implement that TODO-timeout I guess

.ado/pipelines/config/loadtest-baseline-chaos.json
heoelri: i've created these metrics via the portal and then extracted them from there.. i modified some of them manually and that works just fine.. i currently assume that they've to be unique per test..

.ado/pipelines/config/loadtest-baseline-normal.json
heoelri: made our sample guids unique

.ado/pipelines/config/loadtest-baseline-chaos.json
sebader: ok

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -63,25 +68,46 @@ stages:
...
+    - ${{ if eq(parameters.runChaosTesting, 'true') }}:
+      - task: PowerShell@2
+        displayName: 'Override loadTestBaseline for Chaos Experiments'
+        inputs:
+          targetType: inline
+          script: |
+            echo "*** Setting loadTestBaseline to $(chaosLoadTestBaseline)"
+            echo "##vso[task.setvariable variable=loadTestBaseline]$(chaosLoadTestBaseline)"
+
+    # setting parameters for embedded load tests
```
heoelri, line 92:
```suggestion
    # setting parameters for standalone load test runs
```

.ado/pipelines/templates/stages-loadtest-azure.yaml
heoelri: yes, that is correct but i've found no other option to visualize it in the pipeline

this is the same approach we've used for locust - and yes it would also continue if something else breaks

.ado/pipelines/templates/stages-loadtest-locust.yaml
heoelri: will be part of the next PR probably next week

sebader (approved):
nice work! really cool integration with the test criteria

heoelri:
Env needs to be destroyed before the branch can be deleted.

PR #537, opened 2022-08-03
Azure Dashboard - change time range default to 3 days

sebader:
I realized most of the time when we use the dashboard to show anything at all, its the INT one and there we need to switch to show a couple of days worth of data.
So, just proposing to change the default here.


msimecek (approved):
Good idea.

PR #539, opened 2022-08-05
Fix AKS deployment deletion

sebader:
Somehow az aks list seems to be a bit more reliable in catching the clusters. Not sure why, though..

Also added explicit deletion of nginx as part of this.


heoelri (approved):
looks good to me!

PR #540, opened 2022-08-05
Streamlined Cosmos DB dependency logging

msimecek:
Refactoring of the custom dependency logging for Cosmos DB operations. Interesed to see what especially @sebader thinks!

- Using custom `RequestHandler` which is triggered every time the SDK executes request to Cosmos.
  - Automatically adds RU cost and other metadata.
  - Significantly simplifies the code in `CosmosDbService`.
- Exceptions are now thrown and handled correctly.
  - In the `GetX` methods the only thing that can throw exception is JSON parsing, because we mostly use streaming to get items from Cosmos. So the `try-catch` block covers only that and the exception handling doesn't throw duplicate `AlwaysOnDependencyException`s anymore. 

**The main change is what information is now populated to the dependency telemetry.**

image

- **Name** has changed to reflect the name of the calling method.
- **Data** property, which translates to **Command** in AI now contains path to the manipulated item.
- **PartitionKey** is now added as a custom property (making it easier to search, if we wanted).
- **Duration** is populated automatically and reflects the acutal duration of the Cosmos DB request.
  - **ClientElapsedTime** is then added as custom measurement (and is equal to what we used before as **Duration*).
- `success` has special handling for NotFound and Conflict, because we don't want these to show up as errors. Explained in code comment.
- The `ConflictOnInsert` custom property is gone, because the `Conflict` error code is indicative enough.
- Queries which return `FeedIterators` (e.g. `ListCatalogItemsAsync`):
  - Every use of `iterator.ReadNextAsync()` is **tracked as separate dependency call by the RequestHandler**. That means that it's not keeping track of total RU consumption and total number of iterations for queries which return more "pages" of results. 
  - This is a convenience feature - seeing the total cost directly, without the need to sum it with KQL.
  - It was replaced by additional `LogInformation` which shows up as Trace in the request.


src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace AlwaysOn.Shared.TelemetryExtensions
+{
+    public class AppInsightsRequestHandler : RequestHandler
```
sebader, line 13: can we rename that to include "CosmosDb" in the name somehow? Otherwise looking just at the name, its really not clear what kind of request handler this is

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            _telemetryClient = telemetryClient;
+        }
+
+        private readonly TelemetryClient _telemetryClient;
+
+        public override async Task<ResponseMessage> SendAsync(RequestMessage request, CancellationToken cancellationToken)
+        {
+            var title = request.Properties.ContainsKey("Operation") ? request.Properties["Operation"].ToString() : $"{request.Method} {request.RequestUri.OriginalString}";
+            var dbClientEndpoint = request.Properties.ContainsKey("DbClientEndpoint") ? request.Properties["DbClientEndpoint"].ToString() : null;
```
sebader, line 25: doesnt this work?

    request.Properties["DbClientEndpoint"]?.ToString()

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+            //    OK - Target = diagnostics != null ? diagnostics.GetContactedRegions().FirstOrDefault().uri?.Host : _dbClient.Endpoint.Host,
+            //    REMOVED - Success = success
+            //};
+            //
+            //if (response != null)
+            //    OK - telemetry.Metrics.Add("CosmosDbRequestUnits", response.RequestCharge);
+
+            var response = await base.SendAsync(request, cancellationToken);
+
+            dependency.Telemetry.Type = "Azure DocumentDB";
```
sebader, line 48: wasnt there a constant for this before somewhere? or you got rid of that intentionally?

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+            dependency.Telemetry.Data = request.RequestUri.OriginalString; // Will be shown as "Command" in Application Insights
+            dependency.Telemetry.Target = response.Diagnostics != null ? response.Diagnostics.GetContactedRegions().FirstOrDefault().uri?.Host : dbClientEndpoint;
+            dependency.Telemetry.ResultCode = ((int)response.StatusCode).ToString();
+            dependency.Telemetry.Success = response.IsSuccessStatusCode;
```
sebader, line 52: Is a 404 still a success?

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            if (request.Headers.TryGetValue("x-ms-documentdb-partitionkey", out string partitionKey))
+            {
+                dependency.Telemetry.Properties.Add("PartitionKey", partitionKey);
+            }
+
+            return response;
+        }
+
+        public static T CreateOptionsWithOperation<T>(string operationName, string dbClientEndpoint = null) where T : RequestOptions
```
sebader, line 65: can you add a comment what this method does?

src/app/AlwaysOn.Shared/Services/CosmosDbService.cs
```diff
@@ -178,204 +181,123 @@ public async Task DeleteItemAsync<T>(string objectId, string partitionKey)
...
                 // Item stream operations do not throw exceptions for better performance
                 if (responseMessage.IsSuccessStatusCode)
                 {
-                    var item = await JsonSerializer.DeserializeAsync<CatalogItem>(responseMessage.Content, Globals.JsonSerializerOptions);
-                    success = true;
-                    return item;
+                    try
+                    {
+                        var item = await JsonSerializer.DeserializeAsync<CatalogItem>(responseMessage.Content, Globals.JsonSerializerOptions);
+                        return item;
```
sebader, line 199: why not directly return on the previous line?

src/app/AlwaysOn.Shared/Services/CosmosDbService.cs
```diff
@@ -582,162 +450,74 @@ public async Task<IEnumerable<ItemComment>> GetCommentsForCatalogItemAsync(Guid
 
         public async Task AddNewCommentAsync(ItemComment comment)
         {
-            var startTime = DateTime.UtcNow;
-            ItemResponse<ItemComment> response = null;
-            CosmosDiagnostics diagnostics = null;
-            var success = false;
-            var conflict = false;
+            var requestOptions = AppInsightsRequestHandler.CreateOptionsWithOperation<ItemRequestOptions>(nameof(AddNewCommentAsync), _dbClient.Endpoint.Host);
```
sebader, line 453: isnt line basically always the same, with the exception of the method name? can this somehow be simplified?

sebader (changes requested):

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
msimecek: I was thinking about that, but since `RequestHandler` is coming from Cosmos DB I decided to keep it simple. But ofc can do.

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
msimecek: Nope -> `KeyNotFoundException`.

image

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
msimecek: There was, but now this is the only place where this is used, so there's no real need to have a constant for it.

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
msimecek: It's not (as described in the PR description ;) ). I don't remember why we made it a succes, because it shouldn't be - trying to get a non existent object is suspicious.

src/app/AlwaysOn.Shared/Services/CosmosDbService.cs
msimecek: Good catch - I managed to do it elsewhere, but missed here :D

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
sebader: ah too bad. I thought that would work

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
sebader: Hmm I'm not sure I would count a 404 as a non-success. Try to GET item /123 which just doesnt exist (anymore?) is still a successful operation against cosmos. I would have said not success=false means something actually went wrong

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
sebader: Yeah but you need to run Intellisense etc to see where RequestHandler is coming from. So I'd opt to rename it

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
msimecek: But it is actually wrong. The client is trying to get something which is not there.

But fine - I could simply create special case just for 404 in the handler. The other one was Conflict. Do we consider that a success too?

src/app/AlwaysOn.Shared/Services/CosmosDbService.cs
msimecek: Ideas? One possible simplification would be to move this method from App Insights extensions to `CosmosDbService` (as it's the only place where it's used anyway) - therefore having access to _dbClient directly.

Besides that I'm not keen to try and use some reflection magic to populate the method name automatically :) Also it gives you the option to change the operation name to something else.

msimecek: I refactored it, need to test in e2e now :D

msimecek:
Tested each operation individually from localhost. All show up correctly in AI.

msimecek:
e2e: https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=1854

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public override async Task<ResponseMessage> SendAsync(RequestMessage request, CancellationToken cancellationToken)
+        {
+            var title = request.Properties.ContainsKey("Operation") ? request.Properties["Operation"].ToString() : $"{request.Method} {request.RequestUri.OriginalString}";
+            var dbClientEndpoint = request.Properties.ContainsKey("DbClientEndpoint") ? request.Properties["DbClientEndpoint"].ToString() : null;
+
+            using var dependency = _telemetryClient.StartOperation<DependencyTelemetry>(title);
+
+            //
+            // Comparison with the original telemetry object:
```
sebader: lets remove this now

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+            dependency.Telemetry.Type = "Azure DocumentDB";
+            dependency.Telemetry.Data = request.RequestUri.OriginalString; // Will be shown as "Command" in Application Insights
+            dependency.Telemetry.Target = response.Diagnostics != null ? response.Diagnostics.GetContactedRegions().FirstOrDefault().uri?.Host : dbClientEndpoint;
+            dependency.Telemetry.ResultCode = ((int)response.StatusCode).ToString();
+            dependency.Telemetry.Success = success;
+
+            dependency.Telemetry.Metrics.Add("CosmosDbRequestUnits", response.Headers.RequestCharge);
+            dependency.Telemetry.Metrics.Add("ClientElapsedTime", response.Diagnostics != null ? response.Diagnostics.GetClientElapsedTime().TotalMilliseconds : -1);
+
+            if (request.Headers.TryGetValue("x-ms-documentdb-partitionkey", out string partitionKey))
```
sebader, line 48: I guess you checked if this isnt part of the Diagnostics object?!

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            var success = response.IsSuccessStatusCode;
+            // Application-specific special handling in order to have cleaner Application Insights error reporting. Client will still get proper error response.
+            // There are cases which we don't consider database errors and would create noise in monitoring:
+            //  - Considering NotFound as successful call to Cosmos DB, looking for a non-existent object.
+            //  - The same with Conflict, which is handled the application code (ignoring the request, because it already exists).
+            if (response.StatusCode == System.Net.HttpStatusCode.NotFound || response.StatusCode == System.Net.HttpStatusCode.Conflict)
+                success = true;
+
+            dependency.Telemetry.Type = "Azure DocumentDB";
```
sebader, line 39: I wonder if we should add a try-catch around the tracing. just to be very sure. Otherwise it could happen that the request itself worked, but an exception somewhere down here would be thrown up and the caller would assume the operation to cosmos failed?! not sure if something still can go wrong, but I have seen NREs with the Diagnostics before...

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace AlwaysOn.Shared.TelemetryExtensions
+{
+    public class AppInsightsCosmosRequestHandler : RequestHandler
```
sebader, line 13: how about a small comment on class level what this is doing?

sebader (changes requested):

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
msimecek: Yep and it isn't.

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
msimecek: Good point, we probably want the response to be returned in any case :D

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
sebader: you added it on the constructor now. Shouldnt it be on the class?

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
msimecek: Updated

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
msimecek: ..... 🤦

src/app/AlwaysOn.Shared/TelemetryExtensions/AppInsightsRequestHandler.cs
sebader: great

sebader (approved):
great! Lets ship it! :)
also to Connected please

PR #541, opened 2022-08-10
Include the location header when creating new rating

msimecek:
I noticed that the Ratings API doesn't return the Location header when creating a rating. This PR fixes that.


sebader (approved):
I think I had omitted that because I didnt see any point in fetching a particular rating. But you are right, it should be like this

PR #542, opened 2022-08-10
Troubleshooting: Storage key issue

msimecek:
Weird error I bumped into when debugging the application locally. The Storage key worked in e2e, but not on my machine. After regenerating it several times until it didn't contain `/` everything started to work.


sebader (approved):
sounds more like a .NET specific issue with the /

msimecek:
Yes, it's definitely a .NET issue, and particularly when using appsettings.json locally (it works on AKS with the CSI driver). I didn't articulate it properly in the description.

PR #544, opened 2022-08-11
NODE_OPTIONS not needed anymore

msimecek:
With the change to Node 18, the `--openssl-legacy-provider` is not needed anymore.


sebader (approved):
I just assume you tested this :)

PR #548, opened 2022-08-12
Dynamic region setup for Grafana infrastructure

sebader:
This makes the regional distribution of the Grafana infrastructure fully dynamic. So it can be deployed to only one region or n number of them.


src/infra/monitoring/grafana/terraform/stamps/network.tf
```diff
@@ -1,10 +1,10 @@
-# Deploy foundational networking capabilities such as VNETs; subnets and peerings. 
+# Deploy foundational networking capabilities such as VNETs; subnets and peerings.
 
 module "subnet_addrs" {
   source   = "hashicorp/subnets/cidr"
-  for_each = var.stamps
+  for_each = local.stamps
 
-  base_cidr_block = each.value["vnet_address_space"]
+  base_cidr_block = "10.1.0.0/16"
```
heoelri, line 7: Wouldn't this result in the exact same IP ranges being used in each stamp? Not saying that this is nessecarily a problem.

sebader: yes, but otherwise we need to dynamically calculate them and I didnt bother as it really should not matter since they are all standalone

PR #551, opened 2022-08-17
Increase timeout for nginx deployment

sebader:
Attempt to fix fairly frequent timeouts on ngnix deployment

### DIFF
--- a/.ado/pipelines/templates/jobs-deploy-configuration.yaml
+++ b/.ado/pipelines/templates/jobs-deploy-configuration.yaml
@@ -51,6 +51,7 @@ jobs:
                        --set controller.service.loadBalancerIP="$aksIngressIp" `
                        --set controller.service.annotations."service\.beta\.kubernetes\.io/azure-load-balancer-resource-group"="$aksClusterResourceGroup" `
                        --version "$(ingressNginxVersion)" `
+                       --timeout 15m0s `
                        --wait
 
           if ($LastExitCode -ne 0) {


heoelri:
Where's this timeout coming from?

sebader:
> Where's this timeout coming from?

the default for helm to use on `--wait` is 5min, setting it explicitly increases it

heoelri:
> > Where's this timeout coming from?
> 
> the default for helm to use on `--wait` is 5min, setting it explicitly increases it

Sorry, maybe I wasn't clear enough. Why does it take 5+ minutes for helm to deploy ingress-nginx?

sebader:
> > > Where's this timeout coming from?
> > 
> > 
> > the default for helm to use on `--wait` is 5min, setting it explicitly increases it
> 
> Sorry, maybe I wasn't clear enough. Why does it take 5+ minutes for helm to deploy ingress-nginx?

that is the very good question indeed. I couldnt figure that out. It doesnt happen every time, but every now and then, ever since we added `--wait`. My guess is that sometimes the external LB just takes longer to deploy with the public IP. but just a guess

sebader:
like here https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=1894&view=logs&j=60fd2465-6edf-5d3d-5560-6035804ef959&t=bb0b69b5-d1e9-53a5-8d9e-58d80113de13&s=a534c8fe-ffc3-5b1e-3354-818ad12c4222

heoelri:
> > > > Where's this timeout coming from?
> > > 
> > > 
> > > the default for helm to use on `--wait` is 5min, setting it explicitly increases it
> > 
> > 
> > Sorry, maybe I wasn't clear enough. Why does it take 5+ minutes for helm to deploy ingress-nginx?
> 
> that is the very good question indeed. I couldnt figure that out. It doesnt happen every time, but every now and then, ever since we added `--wait`. My guess is that sometimes the external LB just takes longer to deploy with the public IP. but just a guess

You're right. That's a possible explaination. Looking at INT it usually takes only <=2min for the "Install ingress-nginx.." step to complete the deployment for both stamps.

sebader:
yep, usually its pretty fast. And when it fails, its also mostly just one of the two stamps

heoelri (approved):

PR #552, opened 2022-08-17
Query HealthScore as part as Health Service checks

sebader:
In addition to the direct service pings to Event Hub, Cosmos DB and Storage, this adds a query to the Log Analytics workspace to also take the HealthScore into account.

Challenge is yet that the score seems to be not always super reliable and thus the health service might report unhealthy for a few queries.
image

image

@nielsams


src/infra/workload/releaseunit/modules/stamp/keyvault-secrets.tf
```diff
@@ -16,6 +16,8 @@ locals {
     "CosmosDb-ApiKey"                            = data.azurerm_cosmosdb_account.global.primary_key
     "CosmosDb-DatabaseName"                      = var.cosmosdb_database_name
     "API-Key"                                    = var.api_key
+    "LogAnalytics-WorkspaceId"                   = data.azurerm_log_analytics_workspace.stamp.workspace_id
+    "MANAGED-IDENTITY-CLIENTID"                  = azurerm_kubernetes_cluster.stamp.kubelet_identity.0.client_id
```
heoelri, line 20: Is there any specific reason why we use uppercase for some and camelcase for others?

sebader: nope :) just copy-and-paste... but I can switch all to camel

sebader: fixed

src/infra/workload/releaseunit/modules/stamp/keyvault-secrets.tf
```diff
@@ -10,12 +10,14 @@ locals {
...
-    "APPLICATIONINSIGHTS-ADAPTIVE-SAMPLING"      = var.ai_adaptive_sampling,
+    "ApplicationInsights-Connection-String"      = data.azurerm_application_insights.stamp.connection_string
+    "ApplicationInsights-Adaptive-Sampling"      = var.ai_adaptive_sampling,
     "CosmosDb-Endpoint"                          = data.azurerm_cosmosdb_account.global.endpoint
     "CosmosDb-ApiKey"                            = data.azurerm_cosmosdb_account.global.primary_key
     "CosmosDb-DatabaseName"                      = var.cosmosdb_database_name
-    "API-Key"                                    = var.api_key
+    "Api-Key"                                    = var.api_key
+    "LogAnalytics-WorkspaceId"                   = data.azurerm_log_analytics_workspace.stamp.workspace_id
+    "Managed-Identity-Clientid"                  = azurerm_kubernetes_cluster.stamp.kubelet_identity.0.client_id
```
heoelri, line 20:
```suggestion
    "Managed-Identity-ClientId"                  = azurerm_kubernetes_cluster.stamp.kubelet_identity.0.client_id
```

can we use **I**d instead of **id** please? :D

heoelri (approved):
LGTM

sebader:
@nielsams what do you think, should we merge this in its current state already or rather wait?

nielsams:
Looks very good! My only suggestion would be to also parameterize the Kusto query. That may change a few times as we evolve the model.

sebader:
Ok, so now you can disable individual health checks by adding an app setting to the health service like:
image


Then that service is not getting checked:
image

src/app/AlwaysOn.HealthService/HealthJob.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace AlwaysOn.HealthService
+{
+    /// <summary>
+    /// Background job that performs the health checks on a regular interval and caches the result in memory
+    /// Initial source: https://stackoverflow.com/a/68630985/1537195
+    /// </summary>
+    public class HealthJob : BackgroundService
```
msimecek, line 15: Just for me to understand: There was already caching implemented in the previous version. What's the motivation to move to a background job, when the result is used solely by the controller?

src/app/AlwaysOn.HealthService/README.md
```diff
@@ -85,6 +111,18 @@ var testRating = new ItemRating()
 await AddNewRatingAsync(testRating);
 ```
 
+### Azure Monitor HealthScore Query
+
+The regional Azure Monitor Log Analytics workspace is queried for the latest HealthScore. If that is equal or below a certain threshold (default: 0.5, configurable through `HEALTHSERVICE_AZMONITOR_HEALTHSCORE_THRESHOLD` setting), it is considered unhealthy.
+The query can also be configured (`HEALTHSERVICE_AZMONITOR_HEALTHSCORE_QUERY`) currently it uses the follow KQL query:
```
msimecek, line 117:
```suggestion
The query can also be configured (`HEALTHSERVICE_AZMONITOR_HEALTHSCORE_QUERY`) currently it uses the following KQL query:
```

src/app/AlwaysOn.Shared/Interfaces/IDatabaseService.cs
```diff
@@ -1,13 +1,13 @@
...
 using System.Collections.Generic;
-using System.Threading;
 using System.Threading.Tasks;
-using AlwaysOn.Shared.Models;
-using AlwaysOn.Shared.Models.DataTransfer;
 
 namespace AlwaysOn.Shared.Interfaces
 {
-    public interface IDatabaseService
+    public interface IDatabaseService : IHealthCheck
```
msimecek, line 10: I like!

msimecek (changes requested):

src/app/AlwaysOn.HealthService/ComponentHealthChecks/AzMonitorHealthScoreCheck.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                Response<LogsQueryResult> response = await _logsQueryClient.QueryWorkspaceAsync(
+                    _sysConfig.RegionalLogAnalyticsWorkspaceId,
+                    _sysConfig.HealthServiceAzMonitorHealthScoreQuery,
+                    new QueryTimeRange(TimeSpan.FromMinutes(10)),
+                    cancellationToken: cancellationToken);
+
+                LogsTable table = response.Value.Table;
+
+                foreach (var row in table.Rows) // there will be only one row (take 1)
```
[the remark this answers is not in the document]
sebader: hmm good point. Now that the query is outside in the Config, this makes it even harder to understand. But how else should we do it? currently we just expect that the query must only return one row (or we ignore everything else). any ideas?

src/app/AlwaysOn.HealthService/HealthJob.cs
sebader: a) It does consistent checking, not dependent on any external calling it every x seconds
b) It doesnt require any actual Cache, it just uses static objects which makes it much easier to understand

sebader: I'd say it is a cleaner design this way. Plus, the backends get checked consistently not just when somebody decides to call the health service. This way every call to the health service should be equally fast. Not some calls taking longer because a check is now being executed( because cache might have expired). Also I think on the previous design simulatanious calls could have realized that a cache value expired and trigger new downstream checks in parallel

src/app/AlwaysOn.HealthService/ComponentHealthChecks/AzMonitorHealthScoreCheck.cs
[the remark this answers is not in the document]
msimecek: Is it ignoring other rows? The way I read the code it will iterate through all rows until it finds one which is not healthy (then returns `Unhealthy`), or returns `Healthy` if all rows are above threshold.

src/app/AlwaysOn.HealthService/HealthJob.cs
msimecek: After a call I agree.

src/app/AlwaysOn.HealthService/README.md
```diff
@@ -6,11 +6,15 @@ The HealthService is an app component that is running along other components lik
...
 - It attempts to send a message to Event Hub (the message will be filtered out by the background worker)
 - It looks up a state file on the storage account. This file can be used to turn off a region, even while the other checks are still working ok.
 
-All health check results are cached in memory for a configurable number of seconds (by default 10) so that not every call to the API results in backend calls. While this does add a small potential latency in detecting outages, it also reduces the additional cluster load generated by health checks.
+In addition to the direct "pings" to the downstream services, the HealthService also queries Azure Monitor for the current HealthScore (as defined in the Health Model).
+While there is an ingestion latency with Azure Monitor (typically a few minutes), the HealthScore does also take into account additional signals which are not covered by the pings of the HealthService.
+Hence, the combination of both enables the HealthService to do both, a real-time picture of the most critical dependencies of the API, as well as a holistic view of the entire stamp as defined in the health model.
```
msimecek, line 15: If these are supposed to be another paragraphs, it's missing space.

msimecek, line 15:
```suggestion
Hence, the combination of both enables the HealthService to get a real-time picture of the most critical dependencies of the API, as well as a holistic view of the entire stamp as defined in the health model.
```

msimecek (changes requested):

src/app/AlwaysOn.HealthService/README.md
sebader: fixed

msimecek (approved):
LGTM, thanks for this update!

sebader:
will be migrated to Connected

PR #673, opened 2022-11-09
Bump Azure Load Test data plane API to `2022-06-01-preview`

heoelri:
This PR bumps the used data plane API version to `2022-06-01-preview`. The previously used API version `2021-07-01-preview` is deprecated and not available anymore.

The new API version required some changes to make the embedded LT work again. This includes to wait for the uploaded JMX file to be successfully validated before the LT run can be executed/started. This is implemented via a new (optional) `wait` parameter for `file-upload-to-loadtest.ps1`:

image

Setting this parameter to `$true` will wait till the `validationStatus` is `VALIDATION_SUCCESS`.

### DIFF
--- a/.ado/pipelines/templates/stages-loadtest-azure.yaml
+++ b/.ado/pipelines/templates/stages-loadtest-azure.yaml
@@ -268,7 +268,7 @@ stages:
 
             if ( ($resultUrl) -and ($logsUrl) ) {
               New-Item -Path results -ItemType "directory" -Force
-              echo "*** Download results file from $resultsUrl"
+              echo "*** Download results file from $resultUrl"
               Invoke-WebRequest $resultUrl -OutFile "results/results.zip"
               echo "*** Download logs from $logsUrl"
               Invoke-WebRequest $logsUrl -OutFile "results/logs.zip"
--- a/.ado/pipelines/templates/stages-loadtest-azure.yaml
+++ b/.ado/pipelines/templates/stages-loadtest-azure.yaml
@@ -241,7 +241,7 @@ stages:
                   echo "*** Either logsUrl ($logsUrl) or resultUrl ($resultUrl) is empty. Retry $i/3"
                   Start-Sleep -seconds 15
                 } else {
-                  echo "*** Received logsUrl ($logsUrl)
+                  echo "*** Received logsUrl ($logsUrl)"
                   echo "*** Received resultUrl ($resultUrl)"
                 }
 
--- a/.ado/pipelines/templates/stages-loadtest-azure.yaml
+++ b/.ado/pipelines/templates/stages-loadtest-azure.yaml
@@ -212,13 +212,21 @@ stages:
 
                 # processing test data and publish it in azure devops
                 $resultUrl = ($result).testArtifacts.outputArtifacts.resultUrl.url
-                $logsUrl = ($result).testArtifacts.outputArtifacts.logsUrl.url
-                New-Item -Path results -ItemType "directory"
-                echo "*** Download resultsUrl"
-                Invoke-WebRequest $resultUrl -OutFile "results/results.zip"
-                echo "*** Download logsUrl"
-                Invoke-WebRequest $logsUrl -OutFile "results/logs.zip"
 
+                $logsUrl = ($result).testArtifacts.outputArtifacts.logsUrl.url
+
+                # download results and logs
+                if ( ($resultUrl) -and ($logsUrl) ) {
+                  New-Item -Path results -ItemType "directory"
+                  echo "*** Download resultsUrl"
+                  Invoke-WebRequest $resultUrl -OutFile "results/results.zip"
+                  echo "*** Download logsUrl"
+                  Invoke-WebRequest $logsUrl -OutFile "results/logs.zip"
+                } else {
+                  throw "*** ERROR: Either logsUrl ($logsUrl) or resultUrl ($resultUrl) is empty."
+                }
+
+                # throw an error when testResult is FAILED
                 if ($result.testResult -eq "FAILED") {
                   throw "*** ERROR: Test result for run $($result.testRunId) is 'FAILED'. Test did not match the defined test criteria."
                 }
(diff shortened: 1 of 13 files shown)


heoelri:
Successfully validated in E2E, here: https://dev.azure.com/Mission-Critical/Online/_build/results?buildId=2339&view=logs&s=c028e070-ff43-5a93-50d7-93e8da49bd99&j=796565d4-cc10-5120-cbdc-39c8ffce5cb8

src/testing/loadtest-azure/scripts/file-upload-to-loadtest.ps1
```diff
@@ -28,22 +34,50 @@ if (!(Test-Path $testFileName -PathType leaf)) {
...
+
+  do {
+
+    $fileStatus = (& $PSScriptRoot\loadtest-get-files.ps1 -apiEndpoint $apiEndpoint `
+                            -loadTestId $loadTestId `
+                            -fileId $($result.fileId) `
+                            -keepToken $true)
+    if ($fileStatus.validationStatus -ne "VALIDATION_SUCCESS") {
+      Write-Verbose "*** Waiting another 30s for file validation to complete $($fileStatus.validationStatus)"
+      Start-Sleep -seconds 30
```
sebader, line 73: do you have any experience how long the validation usually takes? do we have to wait for 30 or is less also ok?

src/testing/loadtest-azure/scripts/file-upload-to-loadtest.ps1
```diff
@@ -28,22 +34,50 @@ if (!(Test-Path $testFileName -PathType leaf)) {
...
+    } else {
+       Write-Verbose "*** File $($fileStatus.fileId) was successfully validated."
+    }
+
+  } while ($fileStatus.validationStatus -ne "VALIDATION_SUCCESS" )
```
sebader, line 78: what happens if validation fails? that needs to be handled to stop our execution and not wait forever

sebader (changes requested):

src/testing/loadtest-azure/scripts/file-upload-to-loadtest.ps1
heoelri: I've no experience. It's new - there was no need to wait in the previous api version. As of now it looks like it takes only seconds (so most of the time it works w/o going through the loop) but sometimes it takes a little longer.

src/testing/loadtest-azure/scripts/file-upload-to-loadtest.ps1
heoelri: That did not happen so far. But yeah, good catch. Let's see what the other status codes are.

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -162,8 +164,10 @@ stages:
           arguments:
             -loadTestId "$(loadTestId)" `
             -apiEndpoint "$(azureLoadTestDataPlaneURI)" `
+            -apiVersion "$(azureLoadTestApiVersion)" `
             -testFileName "$(System.DefaultWorkingDirectory)/src/testing/loadtest-azure/scripts/catalog-test.jmx" `
-            -verbose:$true
+            -verbose:$true `
```
msimecek, line 169: Is there a reason for these two arguments to use the `-arg:value` syntax, instead of `-arg value` like the others? Is it because they're boolean?

src/testing/loadtest-azure/scripts/file-upload-to-loadtest.ps1
```diff
@@ -17,7 +17,13 @@ param
...
   
   # Test File ID is auto-generated when not set (default)
-  [string] $testFileId = (New-Guid).toString()
+  [string] $testFileId = (New-Guid).toString(),
+
+  # if set to true script will set pipeline variables
+  [bool] $pipeline = $false,
+
+  # if set to true script will wait till file was validated
+  [bool] $wait = $false
```
msimecek, line 26: Is there a scenario where we *don't* want to wait for the validation? I would think that it doesn't work without it, does it?

msimecek (changes requested):

src/testing/loadtest-azure/scripts/file-upload-to-loadtest.ps1
heoelri: My thinking was that this is only needed when the file upload is embedded into the pipeline. And it might not be needed when other files (other than filetype 0 for jmx) are uploaded. I haven't verified that yet as we do not upload any other files at the moment. https://learn.microsoft.com/en-us/rest/api/loadtesting/dataplane/test/get-test-file?tabs=HTTP#filetype

msimecek: I see, thanks. How about setting the default to `$true` for now, as it's unclear whether the validation ever doesn't happen?

src/testing/loadtest-azure/scripts/file-upload-to-loadtest.ps1
heoelri: Reduced it to 15 seconds

src/testing/loadtest-azure/scripts/file-upload-to-loadtest.ps1
heoelri: Modified the loop to run while status is VALIDATION_INITIATED

src/testing/loadtest-azure/scripts/file-upload-to-loadtest.ps1
heoelri:
```suggestion
  [bool] $wait = $true
```

ok for me

.ado/pipelines/templates/stages-loadtest-azure.yaml
heoelri: I'd issues with bool parameters in the past. That's why I'm using `-arg:value` instead of `-arg value` for them.

msimecek (approved):
Can't wait to run it!

PR #742, opened 2022-12-06
Fixing the Front Door metrics

heoelri:
Before

image

and after

image


sebader (approved):
well spotted! probably in Connected too

PR #743, opened 2022-12-07
Migrate to workload identities (managed identities) for AKS

sebader:
Removes the need for any key/secret based authentication against
- Cosmos DB
- Storage accounts
- Event Hubs

Also moves the CSI driver for KeyVault to an per-deployment basis.

Only downside right now: workload identities are still in preview, hence I'm not sure if we want to merge this yet.

@heoelri


heoelri:
> Removes the need for any key/secret based authentication against
> 
> * Cosmos DB
> * Storage accounts
> * Event Hubs
> 
> Also moves the CSI driver for KeyVault to an per-deployment basis.
> 
> Only downside right now: workload identities are still in preview, hence I'm not sure if we want to merge this yet.
> 
> @heoelri Heyko Oelrichs FTE

I think we should merge it. The preview is already bettern than what we had before.

heoelri (approved):

Issue #745, opened 2022-12-07
Azure Load testing Pipeline error (standalone mode)


rishabhsaha:
When I try to run the azure load testing pipeline, it succeeds provisioning the azure load test service but fails with the following error int create test step.

![MicrosoftTeams-image (5)](https://user-images.githubusercontent.com/12716801/206274172-1fb4e2a6-4004-416e-bb40-d3390c4f0002.png)

heoelri:
Hi @rishabhsaha, thanks for reporting this issue. I'll take a look.

PR #746, opened 2022-12-15
Fix issue in Azure Load Testing implementation (Issue #745)

heoelri:
This addresses the issue reported in issue #745 by @rishabhsaha.

### DIFF
--- a/.ado/pipelines/templates/stages-loadtest-azure.yaml
+++ b/.ado/pipelines/templates/stages-loadtest-azure.yaml
@@ -56,11 +56,11 @@ stages:
     - template: steps-parse-terraform-output.yaml
       parameters:
         workingDirectory: '${{ parameters.terraformWorkingDirectory }}'
-        displayName: 'Parse MALT deployment outputs'
+        displayName: 'Parse loadtest deployment outputs'
 
     # Set loadTestBaseline variable
     - task: PowerShell@2
-      displayName: 'Set Azure Load Test baseline file'
+      displayName: 'Set load test baseline file'
       inputs:
         targetType: inline
         script: |
@@ -78,7 +78,7 @@ stages:
 
       # Set loadTestTargetFQDN to the Azure Front Door FQDN (when load test is embedded)
       - task: PowerShell@2
-        displayName: 'Set Azure Load Test variables'
+        displayName: 'Set load test target fqdn'
         inputs:
           targetType: inline
           script: |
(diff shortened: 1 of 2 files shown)


heoelri:
Fixed and validated in e2e and standalone.

.ado/pipelines/templates/stages-loadtest-azure.yaml
```diff
@@ -56,7 +56,16 @@ stages:
...
+        displayName: 'Parse loadtest deployment outputs'
+
+    # Set loadTestBaseline variable
+    - task: PowerShell@2
+      displayName: 'Set load test baseline file'
+      inputs:
+        targetType: inline
+        script: |
+          echo "*** Setting loadTestBaseline to $(azureLoadTestBaseline)"
+          echo "##vso[task.setvariable variable=loadTestBaseline]$(azureLoadTestBaseline)"
```
sebader, line 68: why do you need to set the variable? it already is one just with a different name?!

sebader (commented):
just do double check: the actual fix is to move to UK, right?

.ado/pipelines/templates/stages-loadtest-azure.yaml
heoelri: > just to double check: the actual fix is to move to UK, right?

I moved e2e (back) to `uksouth` as chaos is still not available in `swedencentral`. The issue/fix itself is the changed ordering of the steps, moving "setting loadTestBaseline" outside of the `${{ if eq(parameters.embeddedLoadTest, 'true') }}:` expression.

`loadTestBaseline` is used as the input for `loadtest-create.ps1` - in the `configuration.yaml` we specify two options either `azureLoadTestBaseline` or `chaosLoadTestBaseline` and the pipeline defines, based on some expressions which one will be used for `loadTestBaseline `

heoelri: Default is:

https://github.com/Azure/Mission-Critical-Online/blob/a5552ca1f8324ca7659b81754b26efc043550d4f/.ado/pipelines/templates/stages-loadtest-azure.yaml#L61-L68

And when `runChaosTesting` is set to `true` it will be set to:

https://github.com/Azure/Mission-Critical-Online/blob/a5552ca1f8324ca7659b81754b26efc043550d4f/.ado/pipelines/templates/stages-loadtest-azure.yaml#L88-L96

sebader: ah got it, thanks

sebader (approved):

PR #757, opened 2022-12-22
Enable azure_auth for Grafana (supported in Grafana 9.x)

heoelri:
Needed to use the existing Grafana deployment to query Managed Prometheus data.

https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/prometheus-grafana#create-prometheus-data-source-1


heoelri:
@sebader / @nielsams any concerns? Testing it in INT right now. It is enabling another auth mechanism, no further configuration changes. It "should not" introduce any issues and might also be useful to authenticate against other data sources at some point.

sebader (approved):

heoelri:
Thank you!

image

PR #809, opened 2023-01-31
Duplicated paragraph

RicardoNiepel:


heoelri (approved):
You're absolutely right @RicardoNiepel. Thanks for your contribution!

PR #936, opened 2023-04-14
Update Clean-StaleResources.ps1

YRollHid:
Fixing diagnosticSettings list, otherwise the list will endup empty.


sebader (approved):
Thanks again! LGTM

PR #988, opened 2023-06-21
Rewrite missing regions for Azure Monitor workspace

heoelri:
This PR removes the temporary region rewrite for Workload identity that was used to rewrite regions that did not support Workload Identity while in preview. This is not needed anymore. Workload Identity is now available in all relevant regions.

Adding a region rewrite for Azure Monitor workspace (managed prometheus) while not available in all regions, yet.


sebader:
I'm testing a fallback with a map. e2e running now, lets see

sebader (approved):

Issue #1240, opened 2024-01-22
Disable `Allow blob public access` for load generator storage accounts


heoelri:
A couple of storage accounts are getting flagged. This contains the global storage account where this is needed as it hosts images, but also the storage accounts used by the loadgenerator. Is this needed or can we disable public blob access for the loadgenerator storage accounts? 

> **In-Scope** All Storage Account types, except for File Storage.
> **Out-of-Scope** File Storage Account types. 
> **Logic** If Allow Blob public access is disabled, then the control will pass.Note: ​​​​​​​This control will only be evaluated if the provisioning state is marked as 'Succeeded'. 

@sebader thoughts?

sebader:
For the load generators I'm pretty sure that we can disable it without any issues

sebader:
once this is merged, please redeplo the loadgenerator and lets make sure it runs

heoelri:
> once this is merged, please redeploy the loadgenerator and lets make sure it runs

Yes, needed to check if INT completes successfully first.

PR #1241, opened 2024-01-22
Disable public blob access for loadtest storage accounts

heoelri:
Disable public blob access for loadtest storage accounts and some component updates.

See #1240 for more.


heoelri:
Deployment tested in E2E. Loadgenerator wasn't tested though. https://dev.azure.com/mission-critical/Online/_build/results?buildId=3223&view=results

PR #1359, opened 2024-04-25
Component updates 04/24 and host encryption

heoelri:
This PR contains a could of component updates including:

* Terraform `1.8.1`
* Helm `3.14.4`
* ingress-nginx `4.10.0`
* kubernetes `1.29.2`
* ..

It also enables hostencryption for vms/vmss and makes it configurable via the `variables-<env>.tfvars` files. This is a feature that needs to be enabled at subscription-level before. The troubleshooting guide was updated to include that.

https://github.com/Azure/Mission-Critical-Online/blob/f97ae74cc628112b237d97e7a7f4609546aa42d4/src/infra/workload/releaseunit/variables-e2e.tfvars#L5

The recent ingress controller releases also require `allowSnippetAnnotations` to be set to true in `values.helm.yaml` to allow our mod-security snippets like for example in `ingress.yaml`:

https://github.com/Azure/Mission-Critical-Online/blob/f97ae74cc628112b237d97e7a7f4609546aa42d4/src/app/charts/catalogservice/templates/ingress.yaml#L13-L16

Tested in e2e, here: https://dev.azure.com/mission-critical/Online/_build/results?buildId=3336&view=results


heoelri:
K8s `1.29.2` wasn't included in the first e2e test. Running a new one, here: https://dev.azure.com/mission-critical/Online/_build/results?buildId=3336&view=logs&j=9a62b2c6-931b-5ee5-4062-7dadbfc43efc&t=3795b145-dadf-5bd4-2529-ae897b37f3c3

src/infra/workload/releaseunit/stamp.tf
```diff
@@ -55,4 +57,8 @@ module "stamp" {
   alerts_enabled       = var.alerts_enabled
   api_key              = random_password.api_key.result
   ai_adaptive_sampling = var.ai_adaptive_sampling
+
+  depends_on = [ 
+    azurerm_resource_provider_registration.compute
```
sebader, line 62: Didn't you remove that one again?

docs/reference-implementation/Troubleshooting.md
```diff
@@ -74,6 +74,26 @@ Location: SwedenCentral, Current Limit: 100, Current Usage: 96, Additional Requi
...
+```console
+│ Agent Pool Name: "workloadpool"): performing CreateOrUpdate: unexpected status 400 (400 Bad Request) with response: {
+│   "code": "SubscriptionNotEnabledEncryptionAtHost",
+│   "details": null,
+│   "message": "Subscription does not enable EncryptionAtHost.",
+│   "subcode": ""
+│  }
+```
+
+**Description:** Host encryption needs to be enabled at subscription-level when `aks_enable_host_encryption` is set to `true` in `variables-<env>.tf`.
```
sebader, line 86: If we really need to make host encryption a requirement, we'd need to put it into the getting started guide. Otherwise every single customer will likely fail over this. 
Isn't there some way we can enable this for us without enabling it for everybody who would clone the repo?

heoelri: @sebader I completely forgot this PR. I don't see a way to only enable it for us. The auto-registration via TF is also not an option as it would require us to define all resource provider registrations. I'm ok to add it to the getting started guide though. If that's an acceptable solution.

src/infra/workload/globalresources/main.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "3.94.0"
+      version = "3.97.1"
```
sebader, line 5: how about updating TF everywhere to 3.103.1? otherwise we'll have to update right again afterwards :)

heoelri: Still on it. You might have seen that I've already updated some.

heoelri: But I'm not planning to incorporate any of the other dependabot updates here. The first e2e run already went though - need to change it once again with the changed TF params now. https://dev.azure.com/mission-critical/Online/_build/results?buildId=3360

sebader: I assume there was a reason you went to 3.101.0 instead of latest 3.103.1 ;)

